PR #4, opened 2021-09-25
Used a namespaced argo instance for the site applications

beekhof:
Pros
- namespaced argos have reduced capabilities
- better isolation between sites

Cons
- can't see all applications in one place
- need to look up the URL and password for each argo instance


beekhof:
After consultation, we agree that isolation is a good thing, and that Argo is not and should not try to be the one-stop dashboard for patterns.  Therefor we will merge this PR.

PR #7, opened 2021-11-08
Functionality for a list of namespaces for operators

claudiol:
Adding functionality to list namespaces in the **subcriptions:** section of the values-datacenter.yaml or values-factory.yaml/

Example:  
  # TODO: Allow namespace to be a list
  - name: amq-streams
    namespaces:
      - manuela-data-lake-central-kafka-cluster
      - manuela-tst-all
    channel: amq-streams-1.7.x
    csv: amqstreams.v1.7.1

or 
 - name: advanced-cluster-management
   namespace:  open-cluster-management
   channel: release-2.3
   csv: advanced-cluster-management.v2.3.2

### DIFF
--- a/site/templates/subsciptions.yaml
+++ b/site/templates/subsciptions.yaml
@@ -1,0 +1,40 @@
+{{- range .Values.site.subscriptions }}
+{{- $subs := . }}
+{{- $installPlanValue := .installPlanApproval }}
+
+{{- if $subs.namespaces }}
+{{- range .namespaces }}
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: {{ $subs.name }}
+  namespace: {{ default "openshift-operators" . }}
+spec:
+  name: {{ $subs.name }}
+  source: {{ default "redhat-operators" $subs.source }}
+  sourceNamespace: {{ default "openshift-marketplace" $subs.sourceNamespace }}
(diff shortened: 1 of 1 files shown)


site/templates/subsciptions.yaml
```diff
@@ -0,0 +1,40 @@
...
+{{- $installPlanValue := .installPlanApproval }}
+
+{{- if $subs.namespaces }}
+{{- range .namespaces }}
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: {{ $subs.name }}
+  namespace: {{ default "openshift-operators" . }}
+spec:
```
mhjacks, line 12: I'm a little twitchy about some of these defaults (there seems to be a lot of divergence around channel names in particular, and likewise source and sourceNamespace might be a bit of a surprise...but if someone accepts the defaults and they're wrong, they won't get anything installed, and it's not like we can validate it for them.

I'm OK with this (LGTM)

mhjacks (approved):

site/templates/subsciptions.yaml
claudiol: @mhjacks Indeed. The default values for namespace are not ideal.  Not all operators will be deployed/installed under the openshift-operators namespace.  That said, we need to document this and make sure that we explicitly say that if a namespace is not specified that the operator will attempt to install under the openshift-operator namespace.

PR #8, opened 2021-11-09
Enhance compatibility with older versions of git

mhjacks:
Found out in testing and going through the RHPDS process that `git branch --show-current` is not available on git in EL8 - so we perform the equivalent action using a more compatible mechanism instead.


mbaldessari (commented):
lgtm

PR #22, opened 2022-01-07
Protect ourselves from wrong origin repo

mbaldessari:
When origin is set to the upstream repo we end up pointing to a non
existing repo:

  --set main.git.repoURL="https///github.com/hybrid-cloud-patterns/multicloud-gitops.git"

Let's add some code to protect us from this mistake. So when origin
points to the upstream repo and not our own copy:
$ git remote -v
michele git@github.com:mbaldessari/multicloud-gitops.git (fetch)
michele git@github.com:mbaldessari/multicloud-gitops.git (push)
origin  https://github.com/hybrid-cloud-patterns/multicloud-gitops.git (fetch)
origin  https://github.com/hybrid-cloud-patterns/multicloud-gitops.git (push)

We error out sufficiently early:
$ make install
make -f common/Makefile deploy
make[1]: Entering directory '/home/michele/Engineering/cloud-patterns/multicloud-gitops'
ERROR origin points to wrong repo https///github.com/hybrid-cloud-patterns/multicloud-gitops.git
exit 1
make[1]: *** [common/Makefile:43: deploy] Error 1
make[1]: Leaving directory '/home/michele/Engineering/cloud-patterns/multicloud-gitops'
make: *** [Makefile:7: deploy] Error 2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,11 @@ init:
 	git submodule update --init --recursive
 
 deploy:
+# Protect ourselves from the mistake of having origin point at https:///github.com/..., origin must point to our own clone
+ifeq ($(findstring https///,$(TARGET_REPO)),https///)
+	@echo "ERROR origin points to wrong repo $(TARGET_REPO)"
+	exit 1
+endif
 	helm install $(NAME) common/install/ $(HELM_OPTS)
 
 upgrade:


mhjacks:
I definitely agree we don't want to proceed if we have the wrong repo.  But maybe there's a git command or something we could run that could test repo accessibility with the current config?  The 3 /'s seems like a very specific kind of error, and if we're looking for typos that code might not catch all of them.

There's an article I found after this PR encouraged me to consider the problem: https://stackoverflow.com/questions/9610131/how-to-check-the-validity-of-a-remote-git-repository-url.  Maybe if we ran `git ls-remote` on the target repo URL, we could error out if we get a non-zero result.  What do you think?

mbaldessari:
Agreed on /// being too specific. Sounds like a plan, I'll explore a more simpler approach. Let's hold off on this one for a bit

mhjacks:
I still think we're better off running git ls-remote as a validation step.  could it be as straightforward as (untested):

```@@ -33,13 +33,16 @@ test:
 # Test the charts as the pattern would drive them
        @for t in $(CHARTS); do common/scripts/test.sh $$t normal "$(TEST_OPTS) $(PATTERN_OPTS)"; if [ $$? != 0 ]; then exit 1; fi; done
 
+validate-origin:
+       git ls-remote $(TARGET_REPO)
+
 init:
        git submodule update --init --recursive
 
-deploy:
+deploy: validate-origin
        helm install $(NAME) common/install/ $(HELM_OPTS)
 
-upgrade:
+upgrade: validate-origin
        helm upgrade $(NAME) common/install/ $(HELM_OPTS)
 
 uninstall:
```
?

mbaldessari:
Ack, I'll close this one out for the time being

PR #28, opened 2022-01-19
Add localClusterDomain and fix two nits

mhjacks:
This PR adds the definition of localClusterDomain via ACM config policy.  ACM will push the variable so applications can use it in their own scope for things like configmaps.  This will allow multiple clusters to be configured to have the same clusterGroup and different configmap URLs.

The nits:
1) Remove duplicate subciption/subscription YAML file from ACM, eliminating a warning about ACM subscription being defined twice.
2) Add "| quote" to values in ArgoCD applications to make it easier and more straightforward to use booleans and other non-stringy things in helm overrides


claudiol:
Reviewed all the changes with @mhjacks and they all look good.  Discussed the subscriptions.yaml file and that seems to be in line with our original intent.

PR #35, opened 2022-01-26
Remove description due to regression in GitOps 1.4.0

mhjacks:
Remove description as this is not allowed in GitOps 1.4.0.  Expect it back in 1.4.1 but this will unblock development and demos we need to do this week.


day0hero (approved):
@mhjacks LGTM

mhjacks:
GitOps team expects to release 1.4.1 today so we expect to not need this PR.  Closing.

PR #36, opened 2022-01-29
Remove bashism in vault_exec

mbaldessari:
We currently uase "bash -c" inside vault_exec. This only works when
using UBI-based images. Let's move to 'sh -c' to be a bit more robust
in case we're in the presence of the upstream vault images which do not
have bash installed.

Tested a 'vault-init' on UBI images and it correctly worked with no
errors whatsoever in the log.


mhjacks:
LGTM - I generally agree we should avoid bashisms in general

PR #40, opened 2022-02-03
Add conditional hub-certificate-authority template to push HUB CA to managed clusters

mbaldessari:
By default this ACM templates is inactive and will only be activated if
asked explicitely via the .pushHubCA parameter.
It will pull the ca.crt field from the the kube-root-ca.crt ConfigMap on
the hub into a secret on the managed cluster. This will then be used
by the external-secrets pod so it can trust the https://vault-vault.apps.hub-domain...
API endpoint of the vault.

Tested with this change and once enabled via .pushHubCA the
kubernetes-external-secrets pod could correctly connect to the vault
running on the HUB (without this we'd get self-signed certs errors)


mhjacks (approved):
/lgtm

PR #46, opened 2022-02-17
Rename the namespace and serviceaccounts to the name of the new golang-based external secrets operator

mbaldessari:
We're moving to the newer golang-based external secrets operator at https://external-secrets.io/
To be more explicit about our intention we name namespaces and
serviceaccounts golang-external-secrets. Let's rename it in the hub
config as well.

Tested with the other related golang changes and everything worked as
expected.


mhjacks (approved):
lgtm

PR #49, opened 2022-02-23
Run shellcheck on all shell scripts in scripts/

mbaldessari:
A bunch of changes to get shellcheck to not complain on common/ too much.

Currently I have only tested that make vault-init kept working

﻿- Fix bash linting issues
- Fix quoting issue in examples/kustomize-renderer/kustomize
- Fix some shellcheck issues in scripts/secret.sh
- Disable SC2181 shellcheck test
- Drop unused ns and gitops vars
- Quote password var
- Fix a bunch of quoting in scripts/test.sh
- Disable SC3037 in scripts/secret.sh
- Fix SC2166 in scripts/test.sh
- Shellcheck fixes for scripts/vault-token.sh
- Fix most shellcheck warnings in vault-utils.sh
- Disable SC2016 as we know what we're doing
- Fix SC2013: To read lines rather than words, pipe/redirect to a 'while read' loop in vault-utils

### DIFF
--- a/examples/kustomize-renderer/kustomize
+++ b/examples/kustomize-renderer/kustomize
@@ -1,14 +1,14 @@
 #!/bin/bash -x
 
-BASE=`dirname $0`
-if [ $BASE = $PWD ]; then 
+BASE=$(dirname $0)
+if [ "$BASE" = "$PWD" ]; then
     BASE=./
 fi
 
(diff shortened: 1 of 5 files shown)


mbaldessari:
Let's hold off on merging this until I tested a few more things.

PR #50, opened 2022-02-27
Add 'make test' to every common/ PR

mbaldessari:
- Fix test error in clustergroup example
- Fix acm naked example test
- Fix tests/clustergroup-normal.expected.yaml test
- Add initial checking github action on every pull/push

### DIFF
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -1,0 +1,6 @@
+{
+    "default": true,
+    "MD003": false,
+    "MD013": false,
+    "MD033": false
+}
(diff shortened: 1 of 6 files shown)


mbaldessari:
Mmh not sure why the CI run is only being shown on my branch and not here. I need to figure that part out

beekhof:
Nice :)

beekhof:
Maybe the CI bits need to be merged first

mbaldessari:
I think so, I'll just go ahead and merge this. It's just informational for now, so should not be a big deal.

mbaldessari:
Yep that was it

PR #54, opened 2022-03-08
More CI cleanups for common

mbaldessari:
- Find out charts dynamically
- Fix examples values to use imageregistry
- Move all variable definition into a separate Makefile.vars
- Add naked and normal expected.yaml files for examples-kustomize-renderer chart
- Add proper naked+normal golang-secrets expected yaml files
- Add external-secrets dependency chart to golang-external-secrets
- Make sure we fail if helm template errors out
- Add global hubClusterDomain to golang-external-secrets values
- Remove Makefile.toplevel


mbaldessari:
Let's hold off on merging this one, until I discussed it in more detail with Andrew

mbaldessari:
This is basically preparation work so that we can run make tests and make helmlint from outside of common. E.g. https://github.com/mbaldessari/multicloud-gitops/tree/helm-test

PR #55, opened 2022-03-13
Bring back changes to secret.sh from IE

mbaldessari:
Via 465b1c40 in Industrial Edge we updated common/scripts/secret.sh in
the IE repo only.
Let's bring those changes back into common and while we're at it, let's
make the indenting more consistent.


mhjacks:
lgtm

PR #56, opened 2022-03-13
Remove vault-token.sh

mbaldessari:
It is not used anywhere any longer:
$ grep -ir vault-token industrial-edge multicloud-gitops common medical-diagnosis
$


mhjacks (approved):
lgtm

PR #57, opened 2022-03-18
Tighten vault hub policy a bit

mbaldessari:
Leave only the read one for the time being as it seems to be a saner
default. If more granularity is needed customizations are qreuired in
any case.

Tested and I can still correctly access the secret on the config-demo
app on the regional cluster.


mbaldessari:
@mhjacks could you take a look at this one?

PR #58, opened 2022-03-18
Remove submodule's init target

mbaldessari:
Common is not a submodule any longer. No point in keeping this around


mhjacks (approved):
lgtm

PR #60, opened 2022-03-21
Move vault utils to ansible

mbaldessari:
Tested this change on multicloud-gitops and was able to deploy + unseal the vault and load secrets

﻿- Move vault push secrets task into a separate role
- Shuffle files around to keep the ansible tree structure cleaner
- Remove files from their old place
- Initial work on vault_init
- Avoid using all these playbooks
- Always pass absolute paths
- Implement most of vault_init and vault_unseal
- Complete vault_secrets_init
- Do not print unseal keys in log
- Added vault_pki_internal command
- Silence localhost warning
- Move to tags + cleanups
- Fix up metadata
- Clean up README a bit
- Drop dependencies for now
- Set debug to default(False) inside the play directly since we dropped the var
- Error out nicely when secrets key is missing
- Wait for vault NS and vault pod
- Make secrets parsing a little more robust
- Wait for vault ns+pod using until/retries

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -1,4 +1,4 @@
 [defaults]
 display_skipped_hosts=False
 localhost_warning=False
-roles_path=~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:./roles
+roles_path=./roles:~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles


Makefile
```diff
@@ -50,10 +50,10 @@ uninstall:
 	helm uninstall $(NAME)
 
 vault-init:
-	common/scripts/vault-utils.sh vault_init common/pattern-vault.init
+	common/scripts/vault-utils.sh vault_init $(shell readlink -f common/pattern-vault.init)
 
 vault-unseal:
-	common/scripts/vault-utils.sh vault_unseal common/pattern-vault.init
+	common/scripts/vault-utils.sh vault_unseal $(shell readlink -f common/pattern-vault.init)
```
beekhof, line 56: redundant readlink call?

(got to pretend to be useful somehow)

beekhof:
`+558 −339`... I hope we're getting value for that extra 65% 😇

Makefile
mbaldessari: :) I kinda liked having absolute filenames there and I think there was also another reason for me to move to them, but I tested without and it seems to work from outside common and from inside common (thanks to the the common -> . symlink I suppose). But am not married to it, will remove it ;)

mbaldessari:
I still have some polishing work to do:
1. Add an ansible linting/ci step
2. Check that IE can reuse the bits from common/ correctly
3. Add some pre-flight check that verifies the proper ansible modules are installed (add requirements.txt as well and dependencies in the role)

mbaldessari:
1. and 2. are done. For 3. we'll explore some approaches with Lester and Johnny. Let's keep this unmerged until that is done.

Also I'll need to document the roles/tasks properly before we merge

claudiol:
Looked at the Ansible tasks here.  They all look good from my perspective.  I did notice that there's a few conflicts to take care of.  I'll let Marty weight in.

mbaldessari:
Ok so now this PR contains the following:
1. vault-utils ported to ansible
2. file_unseal: true, variable that controls if the unseal keys should be stored locally on file or inside kubernetes in a secret. The latter is so that we can reuse this to unseal the vault from the operator
3. Added idempotency a bit everywhere so the init + unseal + configuration can be run multiple times (with the goal of running this stuff inside the imperative jobs)

mbaldessari:
I still need to:
- Add documentation in the README
- investigate testing frameworks for this
- Figure out the best way to invoke this via the imperative bits: I.e. I want to detect when we're invoked by the operator and use that knowledge to store the unseal keys in kubernetes in that case *and* to ideally avoid the imperative job at all when we're not being run from the operator

mbaldessari:
Let's hold off on merging, I need to fix two minor things

mhjacks:
LGTM. I will be glad to retire the old shell stuff; this looks a lot better

PR #62, opened 2022-03-24
Add a make help target to list all existing targets (common)

mbaldessari:
This improves the UX discoverability
This will give the following output when run from inside common:

  $ make help
  help             This help message
  argosecret       creates the argo secret
  show             show the starting template without installing it
  test             run helm tests
  helmlint         run helm lint
  validate-origin  verify the git origin is available
  deploy           deploys the pattern
  upgrade          runs helm upgrade
  uninstall        runs helm uninstall
  vault-init       inits, unseals and configured the vault
  vault-unseal     unseals the vault
  load-secrets     loads the secrets into the vault

Tested on Linux (bash+zsh) and on Mac OSX (bash+zsh)


day0hero:
LGTM, nice feature add!

PR #82, opened 2022-04-26
Fix quoting in case of multiple secrets per-key

mbaldessari:
When we merged 964e751db (Quote key and value when uploading a secret)
we partially fixed the quoting which was useful in case of multiline
strings, but in doing so we broke multiple secrets because now we do the
following:

  vault kv put 'secret/hub/group1' 'secret1=value1 secret2=value2'

This will end up with a single secret1 having the value of 'value1 secret2=value2'
which is clearly undesirable.

Let's revisit the approach a bit and make sure we quote each value only
so that when we deal with multiple secrets in one group/key we A) fix
this issue and B) work correctly in case of multi-line secrets.

Lacking a nice and simple multi-line quoting mechanism, we resort to
regex_replace called on all values with the multiline flag and the
dotall flagged enabled directly in the regex (?ms) as ansible does not
allow for those flags to be passed directly.

Tested with Ilkka's reproducer:

    secrets:
      ldap:
        bindPassword: foooooooo
        bindDN: "uid=aap_ldap,cn=sysaccounts,cn=etc,dc=cool,dc=lab"
        url: "ldaps://rh-idm-01.cool.lab"
        groupsBaseDN: "cn=groups,cn=accounts,dc=cool,dc=lab"
        usersBaseDN: "cn=users,cn=accounts,dc=cool,dc=lab"
        ca_crt: |
          -----BEGIN CERTIFICATE-----
          MIIEgjCCAuqgAwIBAgIBATANBgkqhkiG9w0BAQsFADAzMREwDwYDVQQKDAhDT09M
          LkxBQjEeMBwGA1UEAwwVQ2VydGlmaWNhdGUgQXV0aG9yaXR5MB4XDTIyMDEwNTEz
          -----END CERTIFICATE-----

And have gotten the following output:

    $ oc exec -it -n vault vault-0 -- bash -c 'vault kv get secret/hub/ldap'
    ...
    ======== Data ========
    Key             Value
    ---             -----
    bindDN          uid=aap_ldap,cn=sysaccounts,cn=etc,dc=cool,dc=lab
    bindPassword    foooooooo
    ca_crt          -----BEGIN CERTIFICATE-----
    MIIEgjCCAuqgAwIBAgIBATANBgkqhkiG9w0BAQsFADAzMREwDwYDVQQKDAhDT09M
    LkxBQjEeMBwGA1UEAwwVQ2VydGlmaWNhdGUgQXV0aG9yaXR5MB4XDTIyMDEwNTEz
    -----END CERTIFICATE-----
    groupsBaseDN    cn=groups,cn=accounts,dc=cool,dc=lab
    url             ldaps://rh-idm-01.cool.lab
    usersBaseDN     cn=users,cn=accounts,dc=cool,dc=lab

Fixes: https://github.com/hybrid-cloud-patterns/multicloud-gitops/issues/62

### DIFF
--- a/scripts/ansible-push-vault-secrets.sh
+++ b/scripts/ansible-push-vault-secrets.sh
@@ -121,13 +121,41 @@
       - not debug | bool
       - vault_status.rc | int > 0
 
-  - name: Debug
-    debug:
-      msg: "vault kv put '{{ vault_path }}/{{ item.key }}' -> '{{ item.value.keys() | zip(item.value.values()) | map('join', '=') | list | join(' ')}}'"
+# The values-secret.yaml file is in the fairly loose form of:
+# secrets:
+#   group1:
+#     key1: value1
+#     key2: value2
+#   group2:
+#     key1: valueA
+#     key4: valueC
+# The above will generate:
+# vault kv put 'secret/hub/group1' key1='value1' key2='value2'
+# vault kv put 'secret/hub/group2' key1='valueA' key4='valueC'
+# Below we loop on the top level keys (group1, group2) and for each
+# sub-top-level key (key1, key2) we create a single 'key1=value1 key2=value2' string
+#
+# Special note is to be given to the regex_replace which is run against each value aka password:
+# A. The need for it is to quote the passwords
+# B. Since it needs to cater to multiline strings (certs) and ansible offers no way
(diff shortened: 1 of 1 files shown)


ikke-t:
man what a nice regexp trickery :D works for me, I tested. Fixes https://github.com/hybrid-cloud-patterns/multicloud-gitops/issues/62

PR #90, opened 2022-06-13
Improve installation resilience for vault

mhjacks:
1) Sometimes vault namespace does not show up for query when ansible-push-vault-secrets is running. This adds retry logic to give it a few tries before erroring out.

2) Move the rename of the pattern-vault.init (password file) until after the first check for vault readiness. This enables us to run make vault-init idempotently (that is - make vault-init will exit successfully if there's a currently running and correctly configured vault)


mbaldessari:
LGTM

PR #91, opened 2022-06-17
Remove SECRETS from the initial helm chart args passing

mbaldessari:
We're using vault + the load-secrets target to inject secrets in it.
Let's not pass the secret values via helm anywhere as this really
defeats the whole purpose of uploading the secrets out of band.


mhjacks (approved):
/lgtm, agreed

PR #98, opened 2022-06-30
WIP stop fetching values files over URL. Use local files from git

mbaldessari:
WIP Stop fetching values files over URL. Use local files from git

Starting with ArgoCD 2.3 (openshift-gitops 1.4.x) [1] we do can use values
files from git directly. This avoids us fetching values files over http
and potentially avoid issue https://issues.redhat.com/browse/GITOPS-2069
"argocd application do not update when values file changed in helm repo"

We do this for all helm applications *except* when an external chart is
used. In such a case argo will unpack the remote chart into its own
folder and in such a case /values-{hub,region-one}.yaml etc. will not be
present. To make sure that we pass our correct values in this external
chart case, we still go back to the old "fetch-over-http" method.

Tested and got an MCG pattern deployed correctly.

[1] https://github.com/argoproj/argo-cd/blob/bb77664b6f0299bb332843bcd2524820c7ba1558/reposerver/repository/repository.go#L674

### DIFF
--- a/acm/templates/policies/application-policies.yaml
+++ b/acm/templates/policies/application-policies.yaml
@@ -41,8 +41,8 @@ spec:
                     helm:
                       ignoreMissingValueFiles: true
                       valueFiles:
-                      - "{{ coalesce .valuesDirectoryURL $.Values.global.valuesDirectoryURL }}/values-global.yaml"
-                      - "{{ coalesce .valuesDirectoryURL $.Values.global.valuesDirectoryURL }}/values-{{ .name }}.yaml"
+                      - "{{ default "" .valuesDirectoryURL }}/values-global.yaml"
+                      - "{{ default "" .valuesDirectoryURL }}/values-{{ .name }}.yaml"
                       parameters:
                       - name: global.repoURL
                         value: $ARGOCD_APP_SOURCE_REPO_URL
@@ -53,7 +53,7 @@ spec:
                       - name: global.pattern
                         value: {{ $.Values.global.pattern }}
                       - name: global.valuesDirectoryURL
-                        value: {{ coalesce .valuesDirectoryURL $.Values.global.valuesDirectoryURL }}
+                        value: {{ default "" .valuesDirectoryURL }}
                       - name: global.hubClusterDomain
                         value: {{ $.Values.global.hubClusterDomain }}
                       - name: global.localClusterDomain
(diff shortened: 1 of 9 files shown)


mbaldessari:
So now it works. And I correctly get MCG deployed and working. What I am not sure about is the following: Will there be any external charts in the future (it's not the case now with vault) that they will need the correct values files to function properly?
I.e. Should we still fetch values files over http when we are in the presence of an external chart?

acm/templates/policies/application-policies.yaml
```diff
@@ -41,8 +41,8 @@ spec:
                     helm:
                       ignoreMissingValueFiles: true
                       valueFiles:
-                      - "{{ coalesce .valuesDirectoryURL $.Values.global.valuesDirectoryURL }}/values-global.yaml"
-                      - "{{ coalesce .valuesDirectoryURL $.Values.global.valuesDirectoryURL }}/values-{{ .name }}.yaml"
+                      - "{{ default "" .valuesDirectoryURL }}/values-global.yaml"
+                      - "{{ default "" .valuesDirectoryURL }}/values-{{ .name }}.yaml"
```
beekhof, line 45: Pretty sure this is not needed/wrong

beekhof (changes requested):

clustergroup/templates/applications.yaml
```diff
@@ -26,8 +26,13 @@ spec:
     helm:
       ignoreMissingValueFiles: true
       valueFiles:
+      {{- if .chart }}
       - "{{ coalesce .valuesDirectoryURL $.Values.global.valuesDirectoryURL }}/values-global.yaml"
       - "{{ coalesce .valuesDirectoryURL $.Values.global.valuesDirectoryURL }}/values-{{ $.Values.clusterGroup.name }}.yaml"
+      {{- else }}
+      - "{{ default "" .valuesDirectoryURL }}/values-global.yaml"
+      - "{{ default "" .valuesDirectoryURL }}/values-{{ $.Values.clusterGroup.name }}.yaml"
+      {{- end }}
```
beekhof, line 35:
```suggestion
      {{- else }}
      - "/values-global.yaml"
      - "/values-{{ $.Values.clusterGroup.name }}.yaml"
      {{- end }}
```

acm/templates/policies/application-policies.yaml
beekhof, line 45:
```suggestion
                      - "/values-global.yaml"
                      - "/values-{{ .name }}.yaml"
```

mbaldessari:
Dropped in favour of #113

PR #103, opened 2022-07-01
Add job to unseal the vault from the cluster

mbaldessari:
This commit adds support for a separate unsealJob CronJob template which is
enabled when when a user sets clusterGroup.unsealVaultInsideCluster to
true and the template is applied to the hub. This makes it so that the
cluster is unsealed directly via cronjob and the vault unseal keys are
stored inside a secret in OCP. Note that this is provided for
convenience only and it is not recommended to store vault's unseal keys
and root login token inside a kubernetes cluster.

Two different design choices were explored and tried:
A) Via some global switch + an additional unseal template (which is what
this review is all about)
If "clusterGroup.unsealVaultInsideCluster" is true we render an additional template
in common/clustergroup/templates/imperative/unsealjob.yaml which then kicks off
the playbook that unseals the vault and stores the unseal keys inside some secret in
ocp (imperative/vaultkeys by default)

B) Via a defined job (disabled by default) like the following inside
values-hub.yaml:
    - name: vault-unseal
      playbook: common/ansible/playbooks/vault/vault.yaml
      tags: "vault_init,vault_unseal,vault_secrets_init"
      extravars:
        - '{"file_unseal": false}'
      disabled: true

In both cases, we need to make sure that 'make vault-init' does not get
run whenever the "unseal the vault inside the cluster" is enabled.
Otherwise we get two things doing the unsealing in parallel and the
second one (which will almost always be the cronjob) will fail forever.

While choice (B) offered some advantages (less duplication, more
explicit), it made the "detect if unsealVaultInsideCluster inside the
Makefile" rather hard and quite error prone.
Mainly for this reason I ended up with with (A).

Tested with clusterGroup.unsealVaultInsideCluster both set to true and
to false and got a correctly configured and set vault in both.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ HUBCLUSTER_APPS_DOMAIN=$(shell oc get ingresses.config/cluster -o jsonpath={.spe
 
 # --set values always take precedence over the contents of -f
 HELM_OPTS=-f values-global.yaml --set main.git.repoURL="$(TARGET_REPO)" --set main.git.revision=$(TARGET_BRANCH) --set global.hubClusterDomain=$(HUBCLUSTER_APPS_DOMAIN)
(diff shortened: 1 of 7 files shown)


mbaldessari:
Did some refactoring and retested everything. I will make the insecurity of it a bit more prominent tomorrow and then remove the WIP

mbaldessari:
Added a small fix around load-secrets and vault-init running from the inside the cluster. I think this one is ready to go. I retested everything and it should be okay

PR #105, opened 2022-07-05
Add finalizers to all applications

mbaldessari:
As suggested by @beekhof via
https://github.com/hybrid-cloud-patterns/industrial-edge/pull/123/ we
add foreground finalizers for all of our argo applications. While we're
at it we replace acm's application policy finalizer from
"argoproj.io/finalizer" to the same foreground one we use in the other
applications: "resources-finalizer.argocd.argoproj.io/foreground"

This should improve our uninstall story in general.


mbaldessari:
So this is exactly what we discussed over the call today and I just tested it with MCG, so I am going ahead and merge this. Scream if you wanted a different approach

PR #107, opened 2022-07-05
Unify install/deploy and upgrade targets

mbaldessari:
Currently we have two make targets:
1. deploy -> runs 'helm install'
2. upgrade -> runs 'helm upgrade'

You are supposed to use 'make install/deploy' the very first time you
install the pattern. And 'make upgrade' when you "upgrade" the pattern,
which in helm-speak, means you have to run it when you change git branch
or any parameter that gets passed to the very first 'helm install' call.

Since helm offers an 'upgrade --install' option, which installs when the
initial helm chart does not exist, whereas calls does a helm upgrade
when the chart already exists [1], we can unify those two targets and
make them one and the same.

Tested with both an initial install and with an upgrade.

[1] From helm's help:

    $ helm upgrade --help |grep -- "-i, --install"
      -i, --install     if a release by this name doesn't already exist, run an install


claudiol (commented):
This is something we have been talking about doing for a while.  Looks like a sensible change to me.

day0hero:
I agree, this makes sense to me!

PR #110, opened 2022-07-06
Let's switch the legacy-install back to install

mbaldessari:
Via 9ac995caa5d281ec23711a4b643829bc1c214cbb
we switched the default deploy target to be operator-based.
Let's make that the 'operator-deploy/operator-install' target instead for a little
while. That way we can test and get some more soaking-time for this
new installation method. We switch 'make deploy' to go back to the
old basic 'helm install' command.

The plan is still eventually to switch to the new operator-based install
way, this just buys us some time and eases testing without breaking
anyone's workflow.

Tested with both 'make install' (current old-helm-based approach) and
'make operator-deploy' (new operator-based approach)


claudiol (approved):
Totally agree with this. Merging.

PR #112, opened 2022-07-06
fix operator kustomize argo

mbaldessari:
- Cleanup configManagementPlugins section in argocd.yaml
- WIP Reference the values files from the local git repo


mbaldessari:
This won't work if somebody really does set the valuesDirectoryURL to something. So we will need to come up with a plan there

mbaldessari:
Closing this. We're likely to go with #113

PR #113, opened 2022-07-07
Drop valuesDirectoryURL entirely

mbaldessari:
valuesDirectoryURL was only needed when argo did not support referencing
the values files inside the pattern's git repo root [1]. I.e. any value
file that starts with '/' is assumed to reference the git repo's root.

The main reason we do this, is because argo tends to cache values files
over http and ignore refresh requests (needs a hard-refresh), so the
expectation of doing gitops and values files over http breaks down
somewhat. [2]

We have three possible options to fix this:
1. We ignore the issue [2] entirely. We keep things as is and we revert
   the operator change [3]. I.e. we always fetch the values files from
   URL (this keeps things as is and we unbreak IE when deployed via the
   operator), but we are more susceptible to the gitops values caching
   issue. We could reevaluate this once
   https://github.com/argoproj/argo-cd/pull/8322/files gets implemented
   (potentially at least).

2. We completely drop valuesDirectoryURL and we always reference local files.
   External charts need to override all the things. This is quite simple, but
   has a bit of an unknown as we do not know how painful this could be with
   certain external charts, especially because overrides won't allow for the
   use of variables. This is what we implement with this commit here.

3. We add a bunch of ifs everywhere and use local files when: A) we're not an
   external chart and B) when valuesDirectoryURL is not defined

[1] https://github.com/argoproj/argo-cd/blob/bb77664b6f0299bb332843bcd2524820c7ba1558/reposerver/repository/repository.go#L674
[2] https://issues.redhat.com/browse/GITOPS-2069
[3] https://github.com/hybrid-cloud-patterns/patterns-operator/commit/ba3c35a15dd5f239c7a7d337747b01d494278429

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ HUBCLUSTER_APPS_DOMAIN=$(shell oc get ingresses.config/cluster -o jsonpath={.spe
 
 # --set values always take precedence over the contents of -f
 HELM_OPTS=-f values-global.yaml --set main.git.repoURL="$(TARGET_REPO)" --set main.git.revision=$(TARGET_BRANCH) --set global.hubClusterDomain=$(HUBCLUSTER_APPS_DOMAIN)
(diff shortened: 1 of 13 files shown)


mbaldessari:
Tested this on multicloudgitops (one hub and one managed cluster) and everything worked.

mbaldessari:
Tested this on Industrial Edge and went over the results with Lester. Also rechecked with Andrew, merging.

PR #115, opened 2022-07-07
Move operator installation in patterns-operator-system namespace

mbaldessari:
The UI installs it by default in the patterns-operator-system namespace.
Let's have UI and make operator-install in sync on this.


beekhof:
NACK. See channel :)

PR #121, opened 2022-07-22
WIP: Extend values-secret parsing to include files section

mhjacks:
This extends values-secret.yaml parsing to allow entries of the form:

```
files:
  key: <filename>
```
Which requires that <filename> exist and be a valid file.

This will add secrets of the form:

secrets/<path>/key file=<file content>

This works by cat'ing the file into a pipe with oc exec.

### DIFF
--- a/ansible/roles/vault_utils/tasks/push_secrets.yaml
+++ b/ansible/roles/vault_utils/tasks/push_secrets.yaml
@@ -145,7 +145,7 @@
   ansible.builtin.set_fact:
     secret_files_vault_cmds: "{{ secret_files_vault_cmds | default({}) | combine({ item.key: vault_cmd}) }}"
   vars:
-    vault_cmd: "cat '{{ item.value }}' | oc exec -n {{ vault_ns }} {{ vault_pod }} -it -- sh -c 'vault kv put {{ vault_path }}/{{ item.key }} file=-'"
+    vault_cmd: "cat '{{ item.value }}' | oc exec -n {{ vault_ns }} {{ vault_pod }} -it -- sh -c 'vault kv put {{ vault_path }}/{{ item.key }} content=-'"
   loop:
     "{{ file_secrets | dict2items }}"
   loop_control:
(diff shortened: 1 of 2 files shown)


ansible/roles/vault_utils/tasks/push_secrets.yaml
```diff
@@ -158,3 +189,12 @@
   loop_control:
     label: "{{ item.key }}"
   when: not debug | default(False) | bool
+
+- name: Add the file secrets to the vault
+  ansible.builtin.shell:
```
mbaldessari, line 194: Maybe just add # noqa command-instead-of-shell so we can remove the warn_list: in ansible-lint?

mbaldessari (commented):
LGTM overall. Two tiny nits if you have time to do another pass

mhjacks:
OK, those are fixed, and thanks for pointing them out.

Big question - it feels a little odd to me that the structure is different for files: and secrets:, but I did it the way I did because there's a fundamentally different structure in play (that is, you can't really have a multi-valued file). On the one hand it might be nice to keep the same structure in both parts of the file, but on the other (and the way I'm proposing) it's diffrerent; because you'd wind up tossing out part of the metadata.

Also, something I was thinking about is that the hardcoded key "file=" might be ambiguous - is it a name or the actual content? So I changed the hardcoded key to "content" because I feel that's more descriptive and not ambiguous. I expect we will talk about this on the team call Monday and I expect this to be part of the discussion on the Imperative Hackfest Lester requested to follow.

mbaldessari:
Good point re content vs file. I agree that content makes more sense. About the structure I think them being different is fine. Something I forgot to bring up which does not impact this review but is probably worth pointing out: With Andrew we had briefly discussed if there was a way to drop the values-secret.yaml format and just go a simple txt file.
I thought of something like:
```
aws:foo:pass:bar
aws:foo2:pass:bar2
config-demo:secret:pass: baz
```
This is really just an idea trying to answer the question "could we load secrets without ansible only using shell and maybe cut/awk etc?" I.e. we split the strings around the first three columns. First split-string is the path, then the second one is the attribute of the path, pass means "plain password" as opposed to "filename" which would do what you are doing here with ansible.
```
aws:foo:pass: bar
aws:foo2:pass:bar2
config-demo:secret:pass: baz
aap:apikey:filename: <filename.txt> 
```

I need to try a POC to see if this is feasible and how horrible, but I just to wanted to mention it in advance of Monday's call ;)

In general this /lgtm

PR #122, opened 2022-07-26
Add support for extraValueFiles section for applications

mhjacks:
This will allow for the addition of an array of valueFiles to be added in addition to the ones clustergroup adds implicitly (specifically values-global.yaml and values-clustergroup-name.yaml).


mhjacks:
An additional change properly defaults both the vault_cmds and secret_files_vault_cmds, and specifies a new assertion that ensures one or both of secrets: or files: is present

claudiol (commented):
All the changes look good in this PR.  PR approved and will merge.

PR #124, opened 2022-07-28
Preliminary proposal for applicationset support

mhjacks:


clustergroup/templates/applications.yaml
```diff
@@ -1,5 +1,81 @@
...
+          valueFiles:
+            - "values.yaml"
+        {{- range .extraValueFiles }}
+            - {{ . | quote }}
+        {{- end }}
+        {{- if .useGeneratorValues }}
+          values: |-
+            {{ `{{ values }}` }}
+        {{- end }}
+          parameters:
```
beekhof, line 62: What about the other values such as `global.pattern` and `global.targetRevision` ?

mhjacks: That's fair; I hadn't considered that. The values: setting here has to be a string; I'm not sure what would happen if we tried to do something super-fancy with, say, multiple YAML docs. One approach would be to add those to the parameters section (which would be super-straightforward).

Do we have  a scenario where we definitely need those parameters?

PR #126, opened 2022-08-15
Add optional mechanism to influence use of implicit pattern global variables

mhjacks:
This is an idea, suggested by @beekhof.

What this mechanism does is allow an (optional) way to use the values we implicitly define in the framework using different parameter names.

So, for example, if you need to use the localClusterDomain in a chart, but it needs to be defined as a different variable name, you can still use it, like so (assuming it declares a `route.domain` variable):

```yaml
mychart:
  name: mychart
  project: hub
    - route.domain
```

Remember any extra variables declared this way are in addition to the "global" parameters, which will be added to the application definition unconditionally.


mbaldessari:
So I think this makes somewhat sense, although I'd need to really try this. Maybe we can hold off merging this until we have an actual real-life use case/user? As it does add quite a few lines

PR #127, opened 2022-08-17
Align the argo policy

mbaldessari:
This should align the argocd rbac policy for namespaced instances to the
same as the main argo instance.

This was initially pushed via 3ee9bb64127081520050e502e497e55056bfa41a
in multicluster-devsecops. Since it is for common/ let's place it in the
proper repo.


mbaldessari:
Set to draft until we test it thoroughly

beekhof:
lgtm

PR #130, opened 2022-08-24
Change conditionals on imperative framework

mhjacks:
The change in structure is needed because there are two things that require the creation of the imperative role/clusterrole/namespace/rbac:

1) A request for an imperative job
2) A request for the insecure vault unseal operation

These may occur together or separately; an unseal without a job in the old scheme would fail because the role/namespace etc was not created for it

### DIFF
--- a/clustergroup/templates/imperative/job.yaml
+++ b/clustergroup/templates/imperative/job.yaml
@@ -1,5 +1,5 @@
-{{/* Only define this if there are any imperativejobs defined */}}
-{{- if gt (len $.Values.clusterGroup.imperative.jobs) 0 -}}
+{{/* Define this if needed (jobs defined */}}
+{{- if (gt (len $.Values.clusterGroup.imperative.jobs) 0) -}}
 ---
 apiVersion: batch/v1
 kind: CronJob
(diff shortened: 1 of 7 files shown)


clustergroup/templates/imperative/job.yaml
```diff
@@ -1,5 +1,5 @@
-{{/* Only define this if there are any imperativejobs defined */}}
-{{- if gt (len $.Values.clusterGroup.imperative.jobs) 0 -}}
+{{/* Define this if needed (jobs defined */}}
```
mbaldessari, line 1: Only if you happen to do another pass for other reasons, it might be nice to align this comment to the one in [clustergroup/templates/imperative/namespace.yaml](https://github.com/hybrid-cloud-patterns/common/pull/130/commits/3ab539360ab91cfd40a09f80cff282e7d6749de2#diff-c10a3be8e6a5d667e38cc2ad48e0bcb27bc5d8f84a887d24be6228e7fc981d0a)

mbaldessari (approved):
LGTM. Thanks for fixing this case! Merge it once you have tested it

PR #132, opened 2022-08-26
Allow ha vault deployments

ruromero:

Detect if vault deployments has more pods and assume `vault-0` as the leader.
Make the other `vault-x` join the raft cluster as followers and finally unseal the followers.

Additionally: Added check for missing serviceAccount token secret

### DIFF
--- a/ansible/roles/vault_utils/tasks/vault_delete.yaml
+++ b/ansible/roles/vault_utils/tasks/vault_delete.yaml
@@ -1,5 +1,6 @@
 ---
 - include_tasks: pre_check.yaml
+- include_tasks: vault_status.yaml
 
 # Note: We do not wait on purpose here as the pod will be recreated
 # anyways and that would race. We are fine with having it gone once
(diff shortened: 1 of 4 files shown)


mbaldessari (changes requested):
Apologies for CI not running right away (I hope I fixed it). Could you fix the warnings from the linters?:
```
Error: fqcn-builtins Use FQCN for builtin actions.
Error: fqcn-builtins Use FQCN for builtin actions.
Error: yaml no new line character at the end of file
Warning: var-spacing Jinja2 variables and filters should have spaces before and after.
```
Also can you try and use something other than json_query() . That function belongs to community.general and when we ran the unsealing tasks from within the cluster (by settings `insecureUnsealVaultInsideCluster: true`) we run it using the `registry.redhat.io/ansible-automation-platform-22/ee-supported-rhel8:latest` container and that one does not have it. (I know this is a bit annoying, but it was either using that container or rolling our own and maintaining it :) )

ruromero:
@mbaldessari sure it was easier to avoid using it than I expected, it should be fine now

mbaldessari:
Awesome, thanks a lot @ruromero! I'll do a smoke test today or tomorrow and then merge it

mbaldessari (approved):

mbaldessari:
I tested this in non-HA and it worked okay, so I'll merge it now. I might silence that "Get Pods" loop a bit as it is a bit verbose, but that's for later ;) Thanks again!

PR #133, opened 2022-08-26
Add mechanism to create both content and b64content versions

mhjacks:
Writing binary files into vault is easier when base64-encoding the file before it goes in. We have no way of knowing if the user wants that content base64-encoded or not though - smaller, text files need not be.

So this patch creates two attributes for a key - one base64 encoded (b64content), one not (content).


mbaldessari:
(Note that if you rebase this PR to the main tip of common, you'd drop these spurious commits from the PR)

I am wondering if we should just always inject everything as base64 and tweak all external secrets to assume that everything is base64-encoded (we have the b64dec filter in external secrets). It's a bit of a bigger patch and crosses some repos, but maybe it would keep things a little simpler+clearer?

mhjacks:
I'm leery of doing that silently at least because people might want to see what they stored in Vault; and because there are other potential uses for Vault outside of Kubernetes secrets. I think it's a "least surprise" violation how things seem to be handled on retrieval from Vault, but I think it would be a bigger least-surprise validation for us to force b64-encoding on every "file"; though maybe we're just leaving a bit of a landmine for people to trip over if they discover the same thing we did.

Still, I think providing both attributes allows both behaviors - we don't really have control of how users/patterns will consume the secrets from vault, though I could also see a point to "fromKey" being a bit of a surprise with both attributes.

mhjacks:
I've thought about it a bit more, and my thinking is still to publish both attributes. I understand we have b64dec, but in my mind, I think users would want at least the option to see what they stored directly. What do you think?

mbaldessari:
Agreed, let's go for this approach for the time being. I cannot think of any other ways (if we added a "format: {binary,text}" attribute, we probably could not beat external secrets into grasping that. I.e. reading the attribute and using b64dec when format is binary)

PR #140, opened 2022-09-01
Remove CSV after uninstalling

ruromero:

After removing the subscription, the CSV is not deleted and this might cause problems in further installations or look confusing in the Openshift console.


mbaldessari:
Could you try and use -o jsonpath? We're trying to actually slim down any of the tools required locally :)

ruromero:
You're right, old habits. Done @mbaldessari

mbaldessari:
Thanks! Merging

PR #141, opened 2022-09-01
Allow configuring clusters with ExternalSecrets

ruromero:

This is a suggested feature that I decided to use for registering new clusters in ArgoCD instead of using ACM.

By creating ExternalSecrets with a specific label in them, ArgoCD can add them to its configuration. This reduces the technical footprint for cases in which ACM is not necessary.

Check the parameters that are now part of the default configuration.

### DIFF
--- a/clustergroup/values.yaml
+++ b/clustergroup/values.yaml
@@ -41,6 +41,15 @@ clusterGroup:
     roleName: imperative-role
     roleYaml: ""
 
+secretStore:
+  name: vault-backend
+  kind: ClusterSecretStore
+
+# Depends on the value of 'vault_hub' ansible variable used 
+# during the installation
+secretsBase:
+  key: secret/data/hub
+
 #  managedClusterGroups:
 #  - name: factory
 #    # repoURL: https://github.com/dagger-refuse-cool/manuela-factory.git
--- a/examples/values-secret.yaml
+++ b/examples/values-secret.yaml
@@ -17,5 +17,21 @@ secrets:
   aws:
     s3Secret: test-secret
 
-
-
+  # The cluster_xxxx pattern is used for creating externalSecrets that
+  # will be used by ArgoCD to detect other clusters.
+  #
+  # Create a service account with enough permissions and extract the token
+  #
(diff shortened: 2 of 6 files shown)


beekhof:
Can we get some docs and/or a demo recording on how to drive/use this in practice?

clustergroup/values.yaml
```diff
@@ -41,6 +41,15 @@ clusterGroup:
...
     roleYaml: ""
 
+secretStore:
+  name: vault-backend
+  kind: ClusterSecretStore
+
+# Depends on the value of 'vault_hub' ansible variable used 
+# during the installation
+secretsBase:
+  key: secret/data/hub/
```
beekhof, line 51: should drop the tailing '/' right?

ruromero: correct, thanks

ruromero:
> Can we get some docs and/or a demo recording on how to drive/use this in practice?

Yes, I'll add some docs to the PR

examples/values-secret.yaml
```diff
@@ -17,5 +17,21 @@ secrets:
...
+  # The cluster_xxxx pattern is used for creating externalSecrets that
+  # will be used by ArgoCD to detect other clusters.
+  #
+  # Create a service account with enough permissions and extract the token
+  #
+  # CLUSTER_TOKEN=$(oc describe secret -n default argocd-external-token | grep 'token:' | awk '{print$2}')
+  # CLUSTER_CA=$(oc extract -n openshift-config cm/kube-root-ca.crt --to=- --keys=ca.crt | base64 | awk '{print}' ORS='')
+  cluster_example:
+    name: example
+    server: https://api.example.openshiftapps.com:6443
```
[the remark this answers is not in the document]
ruromero: I think it would be enough with a script that creates/extracts the token and eventually adds the secret configuration from a template. It's not about the complexity of creating the secret but the playbook would require credentials to connect to the remote cluster.

mbaldessari (approved):
I think this looks good to me. We can refine a few details later (like what permissions the service account needs etc)

PR #142, opened 2022-09-01
Standardize sync-policy handling 

mhjacks:
This effectively deprecates the variable global.options.syncPolicy but this is the only place we were using it, and we weren't checking syncPolicy here either. This standardizes usage without changing behavior.

### DIFF
--- a/clustergroup/templates/applications.yaml
+++ b/clustergroup/templates/applications.yaml
@@ -147,11 +147,13 @@ spec:
   {{- if .ignoreDifferences }}
   ignoreDifferences: {{ .ignoreDifferences | toPrettyJson }}
   {{- end }}
-{{- if eq $.Values.global.options.syncPolicy "Automatic" }}
+  {{- if .syncPolicy }}
+  syncPolicy: {{ .syncPolicy | toPrettyJson }}
+  {{- else }}
   syncPolicy:
     automated: {}
     #  selfHeal: true
-{{- end }}
+  {{- end }}
 ---
 {{- end }}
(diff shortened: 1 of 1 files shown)


mhjacks:
@sauagarw@redhat.com Would this work for you to solve your limit: problem with ACM?

clustergroup/templates/applications.yaml
```diff
@@ -147,11 +147,13 @@ spec:
   {{- if .ignoreDifferences }}
   ignoreDifferences: {{ .ignoreDifferences | toPrettyJson }}
   {{- end }}
-{{- if eq $.Values.global.options.syncPolicy "Automatic" }}
+  {{- if .syncPolicy }}
+  syncPolicy: {{ .syncPolicy | toPrettyJson }}
+  {{- else }}
```
beekhof, line 152: What about `{{ else if eq $.Values.global.options.syncPolicy "Automatic" }}` ?

mhjacks: a) We don't do that anywhere else
b) If that variable _isn't_ set, the resulting config might be invalid/inconsistent in weird ways. This would remove the inconcistency

mhjacks: This also violates one of my personal pet peeves, of having two different variables named the same thing but having different types (string vs. object). I think the global variable needs to go.

mbaldessari:
LGTM, should we also cleanup all of our patterns to remove the global.options.syncPolicy from the various values-*.yaml files?

PR #143, opened 2022-09-02
Create an example blank chart

beekhof:
To point people to when they think creating one is hard


examples/blank/templates/manifest.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: exmaple
```
mbaldessari, line 4: typo

PR #147, opened 2022-09-05
Integrate the ability to specify IIBs as part of the pattern

beekhof:


mbaldessari:
Can we spell out in the git commit message a bit 1) What IIBs are 2) Why I would use them and when/how?
Also what is that test.yaml supposed to do?

PR #149, opened 2022-09-06
Add hashicorp-vault chart in common/

mbaldessari:
This allows us to simplify any values-* file that has to include vault
and not rely on external chart repos.

Note: We add a hashicorp-vault/update-helm-dependency.sh script which
needs to be run after 'helm dependency update hashicorp-vault' is
called. The reason for this is explained in detail in the README.md
file, but the TLDR is that helm has a bug with subcharts and it won't
remove null values in there and so "host: null" breaks the vault route.

Once either the helm bug is fixed or we switch to OCP 4.11 as a minimum
version and to a vault-helm chart that has subdomains, we can drop this
one liner patch.


mhjacks:
I'm not in love with the idea of patching an upstream dependency this way, but given the alternatives, I think this is the best way to solve the problem until the fix in OCP 4.11 can safely be assumed. Also, there is the benefit that this approach does not change any interfaces.

mbaldessari:
And now I have a real-world use case for this ;) Namely the bobbycar folks use ocpDomain variable to point to "apps.cluster.example.com". So instead of having to change all their charts, I can just set:
```
    - ocpDomain
```

mbaldessari:
Blah commented on the wrong PR ;)

PR #152, opened 2022-09-06
Expand pattern globals

mhjacks:
Same change as intended in https://github.com/hybrid-cloud-patterns/common/pull/126 but since I did that on my main branch, things got weird.


mbaldessari:
And now I have a real-world use case for this ;) Namely the bobbycar folks use ocpDomain variable to point to "apps.cluster.example.com". So instead of having to change all their charts, I can just set:
```
    - ocpDomain
```

PR #155, opened 2022-09-13
Track expected differences bewteen naked and normal test outputs

beekhof:


scripts/test.sh
```diff
@@ -17,28 +17,32 @@ if [ $rc -ne 0 ]; then
...
-    rm -f ${OUTPUT}
-fi
 
 if [ $TEST_VARIANT = normal -a $rc = 0 ]; then
     # Another method of finding variables missing from values.yaml, eg.
     # -    name: -datacenter
     # +    name: pattern-name-datacenter
-    # Alas we can't make it fatal because there *should* be some differences
-    diff -u ${TESTDIR}/${name}-naked.expected.yaml ${TESTDIR}/${name}-normal.expected.yaml
+    diff -u ${TESTDIR}/${name}-naked.expected.yaml ${TESTDIR}/${name}-normal.expected.yaml | sed 's/20[0-9][0-9]-[09][0-9].*//' > ${OUTPUT}.variant
```
mbaldessari, line 30: What is that sed doing?

mbaldessari (commented):
So with this we will track the naked, the normal and the diff, correct?

Once we merge something like this all patterns will need to add the new tests

PR #156, opened 2022-09-13
Use dry-run when running tests

mbaldessari:
This should help avoiding surprises when KUBECONFIG is set or
~/.kube/config exists


mbaldessari:
it does not

PR #162, opened 2022-09-14
Implement a per-subscription csv field

mbaldessari:
Currently a user can only get the global.options.useCSV to true
(defaults to false) and once it is set to true *all* subscrptions
will get their 'startingCSV' fields populated with whatever is set in
the 'csv' subscription field. So all subscriptions *must* have a csv
field set.

With this change we can add a per-subscription useCSV toggle which
allows us to do this case by case.
A use-case here is to pass something like the following:

  subscriptions:
    amqstreams-dev:
      name: amq-streams
      namespace: manufacturing-dev
      channel: amq-streams-1.8.x
      csv: amqstreams.v1.8.0
      installPlanApproval: Manual

The above will make sure the startingCSV is amqstreams.v1.8.0 *only* for
the amq-streams operator, and it will make the installation manual.
So once the manual installation is approved we get v1.8.0 installed.

The caveat is that this is really dependent on how each operator
implements the allowed versions and their lifecycle.

Tested it with the above subscription so and correctly got amq-streams
v1.8.0 after approving manually. The other subscriptions without .csv
had no "startingCSV" field in the Subscription object.

Co-Authored-By: Lester Claudio <claudiol@redhat.com>

### DIFF
--- a/clustergroup/templates/subscriptions.yaml
+++ b/clustergroup/templates/subscriptions.yaml
@@ -27,6 +27,8 @@ spec:
   {{- end }}
   {{- if $.Values.global.options.useCSV }}
   startingCSV: {{ $subs.csv }}
+  {{- else if $subs.csv }}
+  startingCSV: {{ $subs.csv }}
   {{- end }}
 ---
 {{- end }}
@@ -53,6 +55,8 @@ spec:
   {{- end }}
   {{- end }}
   {{- if $.Values.global.options.useCSV }}
+  startingCSV: {{ $subs.csv }}
+  {{- else if $subs.csv }}
   startingCSV: {{ $subs.csv }}
   {{- end }}
 ---
--- a/tests/clustergroup-normal.expected.yaml
+++ b/tests/clustergroup-normal.expected.yaml
@@ -766,6 +766,7 @@ spec:
   sourceNamespace: openshift-marketplace
   channel: release-2.4
   installPlanApproval: Automatic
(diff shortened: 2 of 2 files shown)


beekhof (approved):
/lgtm

clustergroup/templates/subscriptions.yaml
```diff
@@ -54,6 +56,8 @@ spec:
   {{- end }}
   {{- if $.Values.global.options.useCSV }}
   startingCSV: {{ $subs.csv }}
+  {{- else if $subs.csv }}
+  startingCSV: {{ $subs.csv }}
```
beekhof, line 60: Academically... I wonder if this would work (or be any clearer)

```suggestion
  {{- if coalesce $subs.csv $.Values.global.options.useCSV }}
  startingCSV:   {{ coalesce $subs.csv $.Values.global.options.useCSV }}
```

beekhof:
```suggestion
  {{- $useCSV := coalesce $subs.csv $.Values.global.options.useCSV }}
  {{- if $useCSV }}
  startingCSV: {{ $useCSV }}
```

mbaldessari: As Lester correctly pointed out to me global.options.useCSV is a bool so we cannot really coalesce it with the per subscription CSV fiels. I'll go ahead and merge it, since it is what he and I tested

PR #163, opened 2022-09-15
Replace per-app cluster overrides with a more consistent and maintainable argo seed

beekhof:
The existing mechanism scales badly from a configuration perspective, requiring a duplicate of each app for every cluster it should be distributed to.

Additionally, it requires the manual merging of site values files into that of the hub - a configuration that requires additional overhead to maintain and is not actively tested.

The new mechanism preserves the app-of-app semantics and allows the existing site values files to be used, unmodified, to deliver application manifests to the managed clusters directly from the hub without an additional agent being present on the managed clusters.


beekhof:
Linter errors relate to the ansible/vault code that is unchanged but this PR

PR #164, opened 2022-09-15
Reduce the complexity of secrets needed to drive the argo push model

beekhof:
/cc @ruromero

### DIFF
--- a/examples/values-secret.yaml
+++ b/examples/values-secret.yaml
@@ -18,20 +18,15 @@ secrets:
     s3Secret: test-secret
 
   # The cluster_xxxx pattern is used for creating externalSecrets that
-  # will be used by ArgoCD to detect other clusters.
+  # will be used by ArgoCD to push manifests to other clusters.
   #
   # Create a service account with enough permissions and extract the token
   #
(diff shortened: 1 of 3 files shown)


ruromero:
@beekhof that is a good improvement, thanks.

BTW can you add this annotation to the `ExternalSecret`?

```
metadata:
  annotations:
    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
```

For a new install, the `ExternalSecret` will be added at the same time or before than the CRDs are installed and the ArgoCD deployment will not progress successfully.

examples/values-secret.yaml
```diff
@@ -18,20 +18,15 @@ secrets:
     s3Secret: test-secret
 
   # The cluster_xxxx pattern is used for creating externalSecrets that
-  # will be used by ArgoCD to detect other clusters.
+  # will be used by ArgoCD to push manifests to other clusters.
   #
```
ruromero, line 22: You might want to update this comment here with the updated procedure

ruromero (approved):
Great improvement

PR #171, opened 2022-10-02
Drop readlink -f usage

mbaldessari:
And replace it with an echo + pwd combo that works both on Linux
and on Mac OSX.

Note that at this point the Mac OSX support is still not working
with ssh auth due to podman not being able to bind mount host dirs into
the container on the podman machine vm (or at least not /private/tmp)

Tested on both Fedora and Mac OSX


mbaldessari:
Merging as the super linter errors have nothing to do with this

PR #172, opened 2022-10-02
Validated Patterns Schema Validation Update

claudiol:
Having a schema for some of our values files will get us:
 - Minimize mistakes in the values files.
 - If there are mistakes we can discover them before deploying a validated pattern.
 - Having a pre-deploy target might be useful to validate the values files and give
   the user to continue or abort the deployment of the pattern.
 - Enforce structure in key values files used by the validated patterns framework.

The schema for the clustergroup chart will apply to any values-*.yaml file since it
is the main chart for the Validated Patterns framework.  The schema includes our
global and main schemas to enable helm to validate these sections that we use in the
values-global.yaml file.

We think that this is a good solution to ensure that users do not add extraneous
properties to the values files and it will, hopefully, help us with our documentation
of the properties required in each of the values files for Validated Patterns.

### DIFF
--- a/clustergroup/values.schema.json
+++ b/clustergroup/values.schema.json
@@ -30,19 +30,6 @@
             ],
             "title": "ValidatedPatterns"
         },
-        "Main": {
-            "type": "object",
-            "additionalProperties": false,
-            "properties": {
-                "clusterGroupName": {
-                    "type": "string"
-                }
-            },
-            "required": [
-                "clusterGroupName"
-            ],
-            "title": "Main"
-        },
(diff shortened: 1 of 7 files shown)


claudiol:
@mbaldessari Looks like the Super Linter is complaining about some other files in common?

mbaldessari:
New ansible-lint made it into the superlinter. Rebase common and it'll work

claudiol:
Ack @mbaldessari . I will do that first thing. Thanks!

PR #182, opened 2022-10-05
App of App syncwaves

day0hero:
To control the flow of application deployments adding a count to the app
of apps pattern should help with deployments. Each application's counter
will +1 from the previous applications and theoretically should wait for
that application to be healthy/synced before deploying.


beekhof:
There are two possible problems:
1. chart N needs something from chart M
2. chart N needs something from chart M *and* chart M needs something from chart N

Scenario #1 is reasonably ok.  I think most people, even without opening clusterGroup would see that M hasn't started syncing yet and try adjusting the order.  Would adding ordering affect any existing patterns?

Would scenario #2 be just as broken without this change?

soukron:
This issue was opened long time ago, and dependencies between apps is something that is a must. In example: we could need to configure Local Storage Operator before configuring Openshift Data Foundation before configuring Quay registry.

PR #187, opened 2022-10-05
Clarify data requirements for values-secret and fix old expanduser bug

mhjacks:
Fixes https://github.com/hybrid-cloud-patterns/common/issues/184.

Also addresses potential problem with validation by calling expanduser on the files value when validating file existence


mhjacks:
Closed in deference to #186 which was posted while I was putting this together. :)

PR #190, opened 2022-10-07
verify local secret template

mbaldessari:
- Add a pattern_dir variable to the playbook
- Add a check a check_missing_secrets parameter to vault_load_secrets
- Add a number of unit tests to verify the new check_missing_secrets functionality
- Add all the code to actually verify secrets against values-secret-template.yaml but keep it disabled
- Fix a couple of linting issues

### DIFF
--- a/ansible/plugins/modules/vault_load_secrets.py
+++ b/ansible/plugins/modules/vault_load_secrets.py
@@ -51,9 +51,9 @@ files.region2:
 """
 
 import base64
-from collections.abc import MutableMapping
 import os
 import subprocess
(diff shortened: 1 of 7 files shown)


mbaldessari:
Since this is still set to false (https://github.com/hybrid-cloud-patterns/common/pull/190/files#diff-47ee1f6bb0de839f20cc64b425ed71f84a22a348b30791c339fcd1abe54b137cR44) I am merging it for the time being. To enable it by default we'll need to add better docs and check that every pattern has a correctly formed values-secret.yaml.template in the patterns' root dir.

PR #191, opened 2022-10-11
Implement global.clusterVersion

mbaldessari:
We add global.clusterVersion variable (OCP major.minor) and pass it on.
Then we add a /values-<global.clusterVersion.yaml> in the helm parameter
list.

This will allow us to have per-ocp version overrides and potentially
avoid having to use git branches more than needed.

Tested on MCG (w/ACM) and observed the correct global.clusterVersion
on both argo instances on hub and spoke. Also the new values includes
looked correct.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,5 +1,13 @@
 # Changes
 
+## October 13, 2022
+
+* Added global.clusterVersion as a new helm variable which represents the OCP
+  Major.Minor cluster version. By default now a user can add a
+  values-<ocpversion>-<clustergroup>.yaml file to have specific cluster version
+  overrides (e.g. values-4.10-hub.yaml). Will need Validated Patterns >= 0.0.6
+  when deploying with the operator.
+
 ## October 4, 2022
 
 * Extended the values-secret.yaml file to support multiple vault paths and re-wrote
--- a/Changes.md
+++ b/Changes.md
@@ -6,7 +6,8 @@
   Major.Minor cluster version. By default now a user can add a
   values-<ocpversion>-<clustergroup>.yaml file to have specific cluster version
   overrides (e.g. values-4.10-hub.yaml). Will need Validated Patterns Operator >= 0.0.6
-  when deploying with the operator.
+  when deploying with the operator. Note: When using the ArgoCD Hub and spoke model,
+  you cannot have spokes with a different version of OCP than the hub.
 
 ## October 4, 2022
 
--- a/Changes.md
+++ b/Changes.md
@@ -5,7 +5,7 @@
 * Added global.clusterVersion as a new helm variable which represents the OCP
   Major.Minor cluster version. By default now a user can add a
   values-<ocpversion>-<clustergroup>.yaml file to have specific cluster version
-  overrides (e.g. values-4.10-hub.yaml). Will need Validated Patterns >= 0.0.6
+  overrides (e.g. values-4.10-hub.yaml). Will need Validated Patterns Operator >= 0.0.6
   when deploying with the operator.
 
 ## October 4, 2022
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,11 @@ TARGET_REPO=$(shell git remote show $(TARGET_ORIGIN) | grep Push | sed -e 's/.*U
 # git branch --show-current is also available as of git 2.22, but we will use this for compatibility
 TARGET_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
 HUBCLUSTER_APPS_DOMAIN=$(shell oc get ingresses.config/cluster -o jsonpath={.spec.domain})
+HUBCLUSTER_VERSION=$(shell oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}' | sed -n -E 's/([0-9]+).([0-9]+).*/\1.\2/p')
 
 # --set values always take precedence over the contents of -f
 HELM_OPTS=-f values-global.yaml --set main.git.repoURL="$(TARGET_REPO)" --set main.git.revision=$(TARGET_BRANCH) \
-	--set global.hubClusterDomain=$(HUBCLUSTER_APPS_DOMAIN) $(TARGET_SITE_OPT)
+	--set global.hubClusterDomain=$(HUBCLUSTER_APPS_DOMAIN) --set global.clusterVersion="$(HUBCLUSTER_VERSION)" $(TARGET_SITE_OPT)
 TEST_OPTS= -f values-global.yaml --set global.repoURL="https://github.com/pattern-clone/mypattern" \
 	--set main.git.repoURL="https://github.com/pattern-clone/mypattern" --set main.git.revision=main --set global.pattern="mypattern" \
 	--set global.namespace="pattern-namespace" --set global.hubClusterDomain=apps.hub.example.com --set global.localClusterDomain=apps.region.example.com --set global.clusterDomain=region.example.com\
--- a/acm/templates/policies/application-policies.yaml
+++ b/acm/templates/policies/application-policies.yaml
@@ -45,9 +45,9 @@ spec:
                       valueFiles:
                       - "/values-global.yaml"
                       - "/values-{{ .name }}.yaml"
-                      {{- if $.Values.global.clusterVersion }}
-                      - "/values-{{ $.Values.global.clusterVersion }}-{{ .name }}.yaml"
-                      {{- end }}
+                      # We cannot use $.Values.global.clusterVersion because that gets resolved to the
+                      # hub's cluster version, whereas we want to include the spoke cluster version
+                      - '/values-{{ `{{ (cat (semver (lookup "operator.openshift.io/v1" "OpenShiftControllerManager" "" "cluster").status.version).Major (semver (lookup "operator.openshift.io/v1" "OpenShiftControllerManager" "" "cluster").status.version).Minor) | replace " " "." }}` }}-{{ .name }}.yaml'
                       {{- range $valueFile := .extraValueFiles }}
                       - {{ $valueFile | quote }}
                       {{- end }}
--- a/acm/templates/policies/application-policies.yaml
+++ b/acm/templates/policies/application-policies.yaml
@@ -45,6 +45,9 @@ spec:
                       valueFiles:
                       - "/values-global.yaml"
                       - "/values-{{ .name }}.yaml"
+                      {{- if $.Values.global.clusterVersion }}
+                      - "/values-{{ $.Values.global.clusterVersion }}-{{ .name }}.yaml"
+                      {{- end }}
                       {{- range $valueFile := .extraValueFiles }}
                       - {{ $valueFile | quote }}
                       {{- end }}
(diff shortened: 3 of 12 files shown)


tests/acm-normal.expected.yaml
```diff
@@ -527,6 +527,7 @@ spec:
                       valueFiles:
                       - "/values-global.yaml"
                       - "/values-acm-edge.yaml"
+                      - "/values-.yaml"
```
beekhof, line 530: Should there be a default?

mbaldessari: I should actually wrap all those "/values-<ocpversion>.yaml" in some { if .global.clusterVersion }, to be on the safer side in case something goes wrong in fetching the clusterVersion

mbaldessari:
Removed the WIP as I retested everything (testing protocol in the commit message). The only thing I am not signing in blood is that there isn't a better/more proper place to programmatically fetch the current cluster version other than `oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}'`. I looked at the ClusterVersion object but it had channels, desired versions but not the exact *current* cluster ocp version. @beekhof WDYT?

mbaldessari:
Tested this with Johnny and it all worked okay. There is one point worth noting though (nothing too vital, but it's good to be aware):
If you put a subscription in values-4.10.yaml this will be forced on *all* clusters. In a hub and spoke model this might be suboptimal. So for that use-case, we have the following options:
A) We add a values-<ocp-version>-<clustergroup>.yaml support after the values-ocpver.yaml file (more specific)
B) We ignore the problem and let the user add different subscriptions unaware of ocp versions in values-hub.yaml and values-region.yaml
C) We extend the subscription object in common/clustergroup to support a 'appliesToClusterGroup: xyz' and the subscription will be applied only if the cluster is part of that clusterGroup

mbaldessari:
I will try and make this /values-<ocpversion>-<clustergroup>.yaml. I almost have it working, chasing a bug still

mbaldessari:
Ok this is now correct and working. I tested it on a 4.10 hub and a 4.9 spoke and correctly got the following on both argos on the hub (cluster-wide and namespaced):
  helm:
    valueFiles:
...
      - /values-4.10-hub.yaml

And the following on the spoke (both cluster-wide and namespaced):
  helm:
    valueFiles:
..
      - /values-4.9-group-one.yaml

day0hero:
> 

Not sure if this is any more elegant, i normallly run `oc get clusterversion` to get the ocp version ... to pull out the string like you are you could do:  `oc get clusterversion/version -o jsonpath='{.status.desired.version}'`

mbaldessari:
> > Removed the WIP as I retested everything (testing protocol in the commit message). The only thing I am not signing in blood is that there isn't a better/more proper place to programmatically fetch the current cluster version other than oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}'. I looked at the ClusterVersion object but it had channels, desired versions but not the exact current cluster ocp version.
> 
> Not sure if this is any more elegant, i normallly run `oc get clusterversion` to get the ocp version ... to pull out the string like you are you could do: `oc get clusterversion/version -o jsonpath='{.status.desired.version}'`

Right but desired is not necessarily the current version? E.g. the desired version might be 4.10.35 and the cluster is at 4.10.30 because the upgrade process has not yet started or is being rolled out?

mbaldessari:
To be sure I just tested this and I can confirm my theory when you click "Upgrade your cluster", right away `oc get clusterversion/version -o jsonpath='{.status.desired.version}'` returns the *future* cluster version whereas the current version is inside `oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}'`

beekhof:
> Removed the WIP as I retested everything (testing protocol in the commit message). The only thing I am not signing in blood is that there isn't a better/more proper place to programmatically fetch the current cluster version other than `oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}'`. I looked at the ClusterVersion object but it had channels, desired versions but not the exact _current_ cluster ocp version. @beekhof WDYT?

Channel would include major and minor though right?

beekhof:
> To be sure I just tested this and I can confirm my theory when you click "Upgrade your cluster", right away `oc get clusterversion/version -o jsonpath='{.status.desired.version}'` returns the _future_ cluster version whereas the current version is inside `oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}'`

It's not immediately clear which one is preferable :-/

Changes.md
```diff
@@ -1,5 +1,13 @@
 # Changes
 
+## October 13, 2022
+
+* Added global.clusterVersion as a new helm variable which represents the OCP
+  Major.Minor cluster version. By default now a user can add a
+  values-<ocpversion>-<clustergroup>.yaml file to have specific cluster version
+  overrides (e.g. values-4.10-hub.yaml). Will need Validated Patterns >= 0.0.6
```
beekhof, line 8:
```suggestion
  overrides (e.g. values-4.10-hub.yaml). Will need Validated Patterns Operator >= 0.0.6
```

acm/templates/policies/application-policies.yaml
```diff
@@ -45,6 +45,9 @@ spec:
                       valueFiles:
                       - "/values-global.yaml"
                       - "/values-{{ .name }}.yaml"
+                      # We cannot use $.Values.global.clusterVersion because that gets resolved to the
+                      # hub's cluster version, whereas we want to include the spoke cluster version
+                      - '/values-{{ `{{ (cat (semver (lookup "operator.openshift.io/v1" "OpenShiftControllerManager" "" "cluster").status.version).Major (semver (lookup "operator.openshift.io/v1" "OpenShiftControllerManager" "" "cluster").status.version).Minor) | replace " " "." }}` }}-{{ .name }}.yaml'
```
beekhof, line 50: maybe s/cat/print/ ?

install/templates/argocd/application.yaml
```diff
@@ -19,6 +19,9 @@ spec:
       valueFiles:
       - "/values-global.yaml"
       - "/values-{{ .Values.main.clusterGroupName }}.yaml"
+      {{- if .Values.global.clusterVersion }}
```
beekhof, line 22:
```suggestion
      # This assumes that the spokes share the same cluster version as the hub.
      # A limitation we are prepared to live with for now
      {{- if .Values.global.clusterVersion }}
```

beekhof (changes requested):
A few minor nits

mbaldessari:
> > Removed the WIP as I retested everything (testing protocol in the commit message). The only thing I am not signing in blood is that there isn't a better/more proper place to programmatically fetch the current cluster version other than `oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}'`. I looked at the ClusterVersion object but it had channels, desired versions but not the exact _current_ cluster ocp version. @beekhof WDYT?
> 
> Channel would include major and minor though right?



> > Removed the WIP as I retested everything (testing protocol in the commit message). The only thing I am not signing in blood is that there isn't a better/more proper place to programmatically fetch the current cluster version other than `oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}'`. I looked at the ClusterVersion object but it had channels, desired versions but not the exact _current_ cluster ocp version. @beekhof WDYT?
> 
> Channel would include major and minor though right?

Not in a decent form at all, I suspect:
  - channels:
    - candidate-4.10
    - candidate-4.11
    - eus-4.10
    - fast-4.10
    - fast-4.11
    - stable-4.10
    - stable-4.11
    image: quay.io/openshift-release-dev/ocp-release@sha256:15832b3fc1eae4e3fafc97350e5d8a4d1be3948a6e5f5e9db8709b092d27e3da
    url: https://access.redhat.com/errata/RHBA-2022:6532
    version: 4.10.33

There are many availableUpdates channels, and they don't seem to be really usable for this use case.

mbaldessari:
> > To be sure I just tested this and I can confirm my theory when you click "Upgrade your cluster", right away `oc get clusterversion/version -o jsonpath='{.status.desired.version}'` returns the _future_ cluster version whereas the current version is inside `oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}'`
> 
> It's not immediately clear which one is preferable :-/

I suspect we want the actual *current* cluster version and *not* the desired one. Here's my reasoning: Imagine I am upgrading from 4.9 to 4.10. Now if I relied on the desired version, instantly argo would start using /values-4.10-hub.yaml, which would be undesirable because the cluster is still very much on 4.9 at this point. So relying on the openshiftcontrollermanager seems preferrable in general?

Makefile
```diff
@@ -10,10 +10,11 @@ TARGET_REPO=$(shell git remote show $(TARGET_ORIGIN) | grep Push | sed -e 's/.*U
 # git branch --show-current is also available as of git 2.22, but we will use this for compatibility
 TARGET_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
 HUBCLUSTER_APPS_DOMAIN=$(shell oc get ingresses.config/cluster -o jsonpath={.spec.domain})
+HUBCLUSTER_VERSION=$(shell oc get OpenShiftControllerManager/cluster -o jsonpath='{.status.version}' | sed -n -E 's/([0-9]+).([0-9]+).*/\1.\2/p')
 
 # --set values always take precedence over the contents of -f
 HELM_OPTS=-f values-global.yaml --set main.git.repoURL="$(TARGET_REPO)" --set main.git.revision=$(TARGET_BRANCH) \
-	--set global.hubClusterDomain=$(HUBCLUSTER_APPS_DOMAIN) $(TARGET_SITE_OPT)
+	--set global.hubClusterDomain=$(HUBCLUSTER_APPS_DOMAIN) --set global.clusterVersion="$(HUBCLUSTER_VERSION)" $(TARGET_SITE_OPT)
```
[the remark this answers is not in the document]
mbaldessari: Yeah I have a local patch where we just make all these things a per legacy-install target only and I shove them at the bottom of the makefile where they are not as annoying. I would not be entirely against dropping the whole legacy install though (after having fought with these things for a while), so I am more than open to rehash that discussion and/or just drop the thing and be done with it

Changes.md
mbaldessari: Oh gawd yes, thanks

install/templates/argocd/application.yaml
mbaldessari: Not sure I follow here. This totally does work with hub and spoke having different ocp versions across the cluster. I tested with 4.10 on the hub and 4.9 on the spoke

acm/templates/policies/application-policies.yaml
mbaldessari: Do ACM templates have a print function though? I don't see it mentioned here https://github.com/stolostron/go-template-utils nor here https://github.com/stolostron/go-template-utils/commit/5466c772c9e434611156298a80f4c49d7a56bd96#diff-599bd7869ad64965d7a512c105d93e3265bdc6150b13f127c8417ced8bf179b1

day0hero:
@mbaldessari 
I tested this patch using values-${OCP_VERSION}-hub.yaml on medical-diagnosis:
- 4.10 - https://github.com/day0hero/medical-diagnosis/tree/patch-191
- 4.11 - https://github.com/day0hero/medical-diagnosis/tree/patch-191-411

This worked as expected. On the 4.11 cluster in values-hub.yaml i left the odf operator at stable-4.10 and the override was applied correctly.

Makefile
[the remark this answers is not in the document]
beekhof: drop it, drop it, drop it :)

acm/templates/policies/application-policies.yaml
mbaldessari: Ok so there is a printf function and I tested a new variant. Thanks for pointing it out!

Makefile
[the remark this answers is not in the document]
mbaldessari: Yeah I think at this point the tradeoff between building your own operator with a top-level change vs maintaining this cruft is just in favour of the former

install/templates/argocd/application.yaml
mbaldessari: Ok added a note in Changes. (Not adding it in this template as I will drop it later on ;)

mbaldessari:
Retested this fully and addressed the last two comments. Merging

PR #196, opened 2022-10-18
Drop legacy-install from common

mbaldessari:
We now only support the installation via the operator-based templates.
This way we avoid duplication and get a certain amount of flexibility
as the operator is in a better place to fetch information from the
cluster (clusterVersion, DNS, etc.)


mbaldessari:
> One thing that, I think, we will lose is the ability to reinstate the initial install-time values that are referenced in the cluster-wide argo on the hub. Imagine we deploy a pattern on ocp 4.9. The operator fetches the clusterversion from an object and passess it down to argo on the hub clusterwide -> namespaced. So far so good. Now the hub gets upgraded to 4.10. The ClusterVersion helm var will stay "behind" at 4.9. In the legacy-install I can just "fix" this by running 'make legacy-install' and the helm upgrade will repropagate those values. I'd have to check what the procedure would be with the operator?

Or maybe nothing is needed because it's late and I am being dense. I suspect the operator should figure out the different clusterVersion and just push down the new clusterVersion value after the upgrade. Anything else would be a bug in the operator, I suppose

beekhof:
> > One thing that, I think, we will lose is the ability to reinstate the initial install-time values that are referenced in the cluster-wide argo on the hub. Imagine we deploy a pattern on ocp 4.9. The operator fetches the clusterversion from an object and passess it down to argo on the hub clusterwide -> namespaced. So far so good. Now the hub gets upgraded to 4.10. The ClusterVersion helm var will stay "behind" at 4.9. In the legacy-install I can just "fix" this by running 'make legacy-install' and the helm upgrade will repropagate those values. I'd have to check what the procedure would be with the operator?
> 
> Or maybe nothing is needed because it's late and I am being dense. I suspect the operator should figure out the different clusterVersion and just push down the new clusterVersion value after the upgrade. Anything else would be a bug in the operator, I suppose

Indeed :)

mbaldessari:
Closing I did more extensive work here https://github.com/hybrid-cloud-patterns/common/pull/213

PR #206, opened 2022-10-26
Remove .Release.name from namespace labels

mhjacks:


mhjacks:
Removing the namespace from the namespaces section of the pattern causes the pattern to encounter a sync failure on trying to install the subscription (since the NS has to exist first, then apparently the operator does some things with it?).

mhjacks:
Another approach to fix this problem (which I've tested) is to override .Release.name in the pattern overrides section. However, I maintain that using helm variables this way (outside of a context where we're explicitly helm installing something) might be inviting oddities like this, since in most places ArgoCD will "helm template" which won't set these variables

mbaldessari:
I am fine with this change. Do we know why we needed/added the label in the first place? Was there a special reason or it just seemed a good idea at the time?

PR #213, opened 2022-11-06
Makefile cleanups plus legacy target removal

mbaldessari:
- Move testing targets after the installation ones
- Move some of the oc calls into the legacy target
- Simplify getting the remote git URL associated to the target origin
- Drop legacy-install from common
- Move comments in the right place
- Clean up PHONY targets in Makefile
- Move all testing-related variables closer to the tests target
- Make sure we error out after the legacy removal


mhjacks (commented):
What happens to the HUBCLUSTER_* variables?

mbaldessari:
> What happens to the HUBCLUSTER_* variables?

They end up being dropped. The reason I did the half-step in between is for bisectability. The reason they are dropped is that the operator does that lookup and passes down the values of those lookup.

mhjacks:
OK. LGTM

PR #214, opened 2022-11-06
Add support for /values-<platform>-<clustergroup>.yaml

mbaldessari:
In the same vein as /values-<ocpversion>-<clustergroup>.yaml support, we
allow for clusterplatform overrides. Values files now look like the
following:
/values-global.yaml, /values-group-one.yaml, /values-AWS-group-one.yaml, /values-4.11-group-one.yaml

Tested with
https://github.com/hybrid-cloud-patterns/patterns-operator/pull/42 and
could get the correct /values file inclusion in cluster-wide argo and on
the namespaced-argo on both the hub and the spoke clusters.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,5 +1,9 @@
 # Changes
 
+## November 6, 2022
+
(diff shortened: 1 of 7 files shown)


beekhof:
Makes sense to me +1

PR #216, opened 2022-11-09
Drop insecureUnsealVaultInsideCluster

mbaldessari:
We always just unseal the vault from the cluster. We simply drop support
of running these commands manually entirely. The advantage is that the
code is a lot simpler and we have a single avenue to do things.

Tested as following:
1. Deployed MCG and observed loading of secrets happening correctly
2. Removed the imperative/vaultkeys secret
3. Deleted the vault-0 pod
4. Observed that the vault-0 pod was restarted and that the vault was
   sealed
5. Observed that the unsealvault cronjob did nothing
6. Restored the vaultkeys secret
7. Observed that the unsealvault cronjob unsealed the vault correctly
   again

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,5 +1,11 @@
 # Changes
 
+## December 9, 2022
+
+* Dropped insecureUnsealVaultInsideCluster (and file_unseal) entirely. Now
+  vault is always unsealed via a cronjob in the cluster. It is recommended to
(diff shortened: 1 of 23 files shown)


mhjacks:
LGTM as far as the code goes. What's driving this change, though? Do we think we're better off this way from a security standpoint?

claudiol:
This makes sense @mbaldessari . Approved.

PR #218, opened 2022-11-22
Implement new values-secret.yaml format

mbaldessari:
- Initial v2 plumbing
- Add module_utils path for shared common code
- Move v1 secret functions in module_utils
- Split out common function in separate file
- Move the secret format implementation in classes
- A bunch of cleanups of unit-testing. Still not fully passing
- Have super-linter pass
- Fix patching the right function
- Start unit testing V2 as well
- Add version check in v1 implementation
- Force get_version() to return a string for the time being
- Pass proper parameters to V2 class
- Initial start of vault policies implementation
- More validation and more V2 testing
- Make sure we check if the vaultPolicy exists
- Make error message clearer
- Add a lot more tests and start implementing the files: section
- Fix black errors
- Cleanups and also call check_missing_secrets only on v1.0
- Fix black errors
- Add a test to cover for empty vaultPrefixes
- Add support and tests for the password generation via policy
- Actually use the proper oc exec call to generate secrets
- Implement file uploading as base64 attribute as well
- Move more logic into _get_file_path
- Large rewrite to unify fields and files and add initial base64 support and tests
- Add base64 support for secrets as well and fix up tests
- POC mostly working at this point
- For the time being always stay on getpass.getpass() for input
- Remove unused modules
- Add checks for duplicate secret names and field names
- Switch to using AnsibleModule.run_command
- Make the secret prompting a little prettier
- Implement looking for ~/values-secret-<patternname>.yaml first
- Test base64 secret
- Silence yaml indentation linting errors and more
- Added values-secret examples


ansible/tests/unit/v2/values-secret-v2-base.yaml
```diff
@@ -0,0 +1,42 @@
...
+    fields:
+    - name: secret
+      onMissingValue: generate # One of: error,generate,prompt
+      vaultPolicy: basicPolicy
+    - name: secret2
+      value: null
+      onMissingValue: prompt
+    - name: ca_crt
+      path: /tmp/ca.crt
+      onMissingValue: error # One of error, prompt (for path). generate makes no sense for file
```
beekhof, line 38: Lets leave out the internal "note to self" comments :)

ansible/tests/unit/v2/values-secret-v2-base.yaml
```diff
@@ -0,0 +1,42 @@
+# Needed to specify the new format (missing version means old version: 1.0 by default)
+version: 2.0
+# NEVER COMMIT THESE VALUES TO GIT
```
beekhof, line 3: Is this true anymore?

examples/secrets/values-secret.v2.yaml
```diff
@@ -0,0 +1,81 @@
...
+  #     secretfile: ...<content of the file as input by user. If just enter is pressed the file will be /tmp/ca.crt...0
+  #     ca_crt: ...<content of /tmp/ca.crt>...
+  #     ca_crt_b64: ...<content of /tmp/ca.crt base64-encoded before uploading to vault>...
+  # - secret/snowflake.blueprints.rhecoeng.com:
+  #     secret: ...<generated basicPolicy secret>...
+  #     secretprompt: ...<as input by the user>...
+  #     secretprompt2: ...<as input by the user. If just enter is pressed it will be 'defaultvalue'>...
+  #     secretfile: ...<content of the file as input by user. If just enter is pressed the file will be /tmp/ca.crt...0
+  #     ca_crt: ...<content of /tmp/ca.crt>...
+  #     ca_crt_b64: ...<content of /tmp/ca.crt base64-encoded before uploading to vault>...
```
beekhof, line 41: We should try to avoid providing documentation in example manifests

examples/secrets/values-secret.v2.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - region-one
+    - snowflake.blueprints.rhecoeng.com
+    fields:
+    - name: secret
+      onMissingValue: generate # One of: error,generate,prompt
+      vaultPolicy: basicPolicy
+    - name: secretprompt
+      value: null
+      onMissingValue: prompt
+      description: "Insert secretprompt NOW"
```
beekhof, line 54: Since people will cargo-cult the examples here, how about:
   "Please specify the (password|API key|...) used by ABC application to access the XYZ service"

examples/secrets/values-secret.v2.yaml
```diff
@@ -0,0 +1,81 @@
...
+  # - secret/snowflake.blueprints.rhecoeng.com:
+  #     secret: ...<generated basicPolicy secret>...
+  #     secretprompt: ...<as input by the user>...
+  #     secretprompt2: ...<as input by the user. If just enter is pressed it will be 'defaultvalue'>...
+  #     secretfile: ...<content of the file as input by user. If just enter is pressed the file will be /tmp/ca.crt...0
+  #     ca_crt: ...<content of /tmp/ca.crt>...
+  #     ca_crt_b64: ...<content of /tmp/ca.crt base64-encoded before uploading to vault>...
+  - name: config-demo
+    vaultMount: secret
+    vaultPrefixes:
```
beekhof, line 44: Can we do some cursory research how (for example) CyberArk structures things and potentially choose terms that are independent of the implementation?

examples/secrets/values-secret.v2.yaml
```diff
@@ -0,0 +1,81 @@
+# NEVER COMMIT THESE VALUES TO GIT (unless your file only uses generated
+# passwords or only points to files)
+
+# Needed to specify the new format (missing version means old version: 1.0 by default)
+version: 2.0
```
beekhof, line 5: Suggest we also include something to the effect of:

`backingStore: vault`

examples/secrets/values-secret.v2.yaml
```diff
@@ -0,0 +1,81 @@
...
+    rule "charset" { charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZ" min-chars = 1 }
+    rule "charset" { charset = "0123456789" min-chars = 1 }
+
+  advancedPolicy: |
+    length=20
+    rule "charset" { charset = "abcdefghijklmnopqrstuvwxyz" min-chars = 1 }
+    rule "charset" { charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZ" min-chars = 1 }
+    rule "charset" { charset = "0123456789" min-chars = 1 }
+    rule "charset" { charset = "!@#$%^&*" min-chars = 1 }
+
```
beekhof, line 24: I wonder if it would be good idea to have a few default (named) policies in the playbook and support:

```
enableDefaultVaultPolicies: 
- basic
- advanced
``` 

to avoid a) having this giant block everywhere, and b) keeping it synchronised if it ever needs to be updated

Changes.md
```diff
@@ -1,5 +1,15 @@
 # Changes
 
+## November 22, 2022
+
+* Implemented a new format for the values-secret.yaml. Example can be found in examples/ folder
+* Now the order of values-secret file lookup is the following:
+  1. ~/values-secret-<patternname>.yaml
+  2. ~/values-secret.yaml
+* Add support for ansible vault encrypted values-secret files. You can now encrypt your values-secret file
+  with `ansible-vault encrypt ~/values-secret.yaml`. The load-secrets mechanism will detect that and prompt
```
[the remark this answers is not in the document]
mbaldessari: No, I need to add more docs to that. But basically you can encrypt your values file with `ansible-vault encrypt` and if you did, `make load-secrets` will figure it out and prompt you for a password if it is encrypted

ansible/tests/unit/v2/values-secret-v2-base.yaml
mbaldessari: Aye, in the unit test files that is a fair point. Will do

ansible/tests/unit/v2/values-secret-v2-base.yaml
mbaldessari: Sometimes? I mean it is true unless the secret file is based on generate or prompt *only*, so I think it is best practice to just not commit it to git

examples/secrets/values-secret.v2.yaml
mbaldessari: Ideas on that? I am thinking of adding a schema json file, and see if we can generate docs from that directly and potentially validate it using it (although the validation might only be very superficial. I.e. you can't do too complex stuff with it)

examples/secrets/values-secret.v2.yaml
mbaldessari: Ack

examples/secrets/values-secret.v2.yaml
mbaldessari: Ack

examples/secrets/values-secret.v2.yaml
mbaldessari: Docs to upload secrets were a bit hidden IMO. https://docs.conjur.org/Latest/en/Content/Developer/Conjur_API_Set_Secret.htm?TocPath=Developer%7CREST%C2%A0APIs%7C_____9 has something. I suspect that the 'path' concept of uploading a secret somewhere is similar. What I am not entirely sure, and could not figure it out from a cursory look at the docs, is, if given a path, you can have multiple attributes with secret material for each attribute.
In any case though I don't know that exposing generic terms that we have then to translate (potentially imperfectly) internally is the way to go as it is bound to be a lossy direction and a lot of code that is bound to have bugs (same with hiding a password policy behind some generic term instead of saying "this is the vaultPolicy")

examples/secrets/values-secret.v2.yaml
mbaldessari: I like this idea. Maybe I can just hard-code a reasonable single validatedPatternDefaultPolicy (complex name chosen so it won't overlap ever with something existing) and we push that all the time and examples are guaranteed that they reference it?

ansible/tests/unit/v2/values-secret-v2-base.yaml
mbaldessari: I removed it from the unit test files, but we will want something in the official examples or in the docs telling the user that in general it is best not to commit to git *unless* all the passwords are generated

examples/secrets/values-secret.v2.yaml
mbaldessari: Ok I added a new hard-coded policy https://github.com/hybrid-cloud-patterns/common/pull/218/commits/36d06e1fccd43f52a57d4d8c56aafa25b4de6ec1#diff-600beb9d1915faccc894389dbc01346727fa651ba20c4212128dca0857351a90R27
I think it is okay to hard-code it as it does nothing if not used. I'd stick to just one as I suspect it will be mainly used for demos and whatnot.

PR #229, opened 2022-11-29
Vault SSL by default

mbaldessari:
- Filter out patches to test/
- Add patch to be able to use vault-ssl as a default
- Switch to vault-ssl as a default
- Fix up tests after ssl as default


mbaldessari:
Still need to test a few things

PR #232, opened 2022-12-14
Drop any remains of output_file

mbaldessari:
This was only used when we had support for storing the unseal keys in
files. Since we dropped this support we can remove any traces of that.


mhjacks:
Yup. Burn it with fire! :fire: /LGTM

PR #239, opened 2023-01-10
Default vaultPrefixes to be ["hub"]

mbaldessari:
This way the secret templates can be less verbose in most cases.


mbaldessari:
Tested on MCG, merging

PR #243, opened 2023-01-11
Add ini-file support

mbaldessari:
- Add a get_ini_value() commodity function
- Fix up ini function and add tests
- Add ini_file support


mhjacks (approved):
LGTM. No new dependencies and I think this will be a major benefit to our users

claudiol (approved):
Changes reviewed.  INI format support will go a long way to offer a way to represent secrets for applications that will be added to Vault.

PR #248, opened 2023-01-16
validate schema makefile

claudiol:
- Adding validate-schema target
- Added entry to Changes.md for validate-schema Makefile target

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,5 +1,9 @@
 # Changes
 
+## January 16, 2023
+
+* Added validate schema target as an option for patterns to use
+
 ## January 13, 2023
 
 * Simplify the secrets paths when using argo hosted sites
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,12 @@ help: ## This help message
 	@printf "$$(grep -hE '^\S+:.*##' $(MAKEFILE_LIST) | sed -e 's/:.*##\s*/:/' -e 's/^\(.\+\):\(.*\)/\\x1b[36m\1\\x1b[m:\2/' | column -c2 -t -s :)\n"
 
 #  Makefiles in the individual patterns should call these targets explicitly
+#  e.g. from industrial-edge: make -f common/Makefile validate-schema
+.PHONY: validate-schema
+validate-schema: ## validate the schema against clustergroup for the pattern before deploying 
+	helm template common/clustergroup/ --name-template $(NAME) $(HELM_OPTS)
+
+#  Makefiles in the individual patterns should call these targets explicitly
 #  e.g. from industrial-edge: make -f common/Makefile show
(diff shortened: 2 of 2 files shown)


mbaldessari:
The reason I am not super fond of adding this is that the helm options that we pass here, are not the ones that argo will pass down. Those have a bunch of additional values-* files whereas here we only have values-global.yaml file. So this would be a very rough approximation. If we were to add this here properly we would have to readd a lot of logic with all the right values files to make it useful?

mbaldessari:
We could do something like:
```
diff --git a/common/Makefile b/common/Makefile
index d95b889..760072c 100644
--- a/common/Makefile
+++ b/common/Makefile
@@ -35,8 +35,13 @@ validate-origin: ## verify the git origin is available
                echo "Running inside a container: Skipping git ssh checks";\
        fi

+.PHONE: validate-schema
+validate-schema: ## validates values files against schema in common/clustergroup
+       $(eval VAL_PARAMS := $(shell for i in ./values-*.yaml; do echo -n "-f $${i} "; done))
+       helm template common/clustergroup --name-template $(NAME) $(HELM_PARAMS) $(VAL_PARAMS) >/dev/null
+
 .PHONY: operator-deploy operator-upgrade
-operator-deploy operator-upgrade: validate-origin ## runs helm install
+operator-deploy operator-upgrade: validate-origin validate-schema ## runs helm install
        helm upgrade --install $(NAME) common/operator-install/ $(HELM_OPTS)

 .PHONY: deploy upgrade legacy-deploy legacy-upgrade
```

But if we do this, I think we should first investigate and fix these warnings as they'd scare people off:
```
helm template common/clustergroup --name-template multicloud-gitops  -f ./values-global.yaml -f ./values-group-one.yaml -f ./values-group-two.yaml -f ./values-hub.yaml  >/dev/null
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/michele/Engineering/Labs/hub-region-aws/region/auth/kubeconfig
coalesce.go:223: warning: destination for pattern-clustergroup.clusterGroup.applications is a table. Ignoring non-table value ([])
coalesce.go:223: warning: destination for pattern-clustergroup.clusterGroup.managedClusterGroups is a table. Ignoring non-table value ([])
coalesce.go:223: warning: destination for pattern-clustergroup.clusterGroup.subscriptions is a table. Ignoring non-table value ([])
make[1]: Leaving directory '/home/michele/Engineering/cloud-patterns/multicloud-gitops'
```

mbaldessari:
Warning fixes are here https://github.com/hybrid-cloud-patterns/common/pull/251

claudiol:
Closing since PR #252 supersedes the changes here.

PR #250, opened 2023-01-17
Extend framework to allow for ESO on ACM-imported clusters

mbaldessari:
This change does three main things:
1. It extends the vault unseal job on the hub with a few tasks that fetch all
   non-hub ACM clusters and for each cluster it connects to the remote
   cluster API endpoint, fetches the golang-external-secret SA token
   and uses it to configure the secret auth/<clusterfqdn>/config
   kubernetes vault backend. If no clusters exist, the additional tasks
   do nothing. Furhtermore, the tasks are also idempotent.

2. It adds a job called acmhub-ca-cronjob on the non-hub clusters which
   will fetch the ACM hub's CA and put it in a secret called "hub" in
   the imperative namespace. Idempotent and noop when the cluster is not
   part of ACM

3. When the golang-external-secrets chart is included on a spoke
   cluster, it will automatically point to the vault on the hub

With this change a spoke cluster may decide to install the
golang-external-secrets chart on a spoke and avoid pushing secrets
around via ACM. By doing so we allow a developer to decide if ESO on the
spokes is wanted and we also do not break the existing way of doing
things.

Tested with https://github.com/mbaldessari/multicloud-gitops/commit/4d90c85149fe0a957f6a69666472a51a9bb37c6b
and was able to get MCG working on HUB and spoke with two different
secrets.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,5 +1,9 @@
 # Changes
 
+## January 23, 2023
+
+* Add initial support for running ESO on ACM-imported clusters
+
 ## January 18, 2023
 
 * Add validate-schema target
--- a/ansible/playbooks/acm/acmhub-get-ca.yaml
+++ b/ansible/playbooks/acm/acmhub-get-ca.yaml
@@ -1,0 +1,52 @@
+# This playbook fetches the hub cluster's CAbundle from ACM's objects
+# and puts it in a secret inside the imperative namespace
+---
+- hosts: localhost
+  connection: local
+  gather_facts: false
+  become: false
+  vars:
+    ns: imperative
+  tasks:
+    - name: Find hub cluster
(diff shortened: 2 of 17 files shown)


claudiol (approved):
Approved. Basically this extends the secrets to the spokes and get their secrets from the Vault on the Hub.  So then we wouldn't have to include an ACM policy to copy the secret over to the Spokes. Looks like a nice alternative to the policy being used in some of our patterns to push the secrets using ACM.

mbaldessari:
Let's wait on the others before merging this

mbaldessari:
Just added an entry in the Changes.md file and retested this to make sure it works when pushing secrets around via ACM with ESO running only on the hub. Discussed with Martin and Andrew as well, merging

PR #264, opened 2023-02-02
Add --pull=newer when running the container

mbaldessari:
From https://docs.podman.io/en/latest/markdown/podman-run.1.html#pull-policy

Pull image policy. The default is missing.

    always: Always pull the image and throw an error if the pull fails.
    missing: Pull the image only if it could not be found in the local
             containers storage. Throw an error if no image could be found and
             the pull fails.
    never: Never pull the image but use the one from the local containers
           storage. Throw an error if no image could be found.
    newer: Pull if the image on the registry is newer than the one in
           the local containers storage. An image is considered to be newer
           when the digests are different. Comparing the time stamps is prone
           to errors. Pull errors are suppressed if a local image was found.

Switching to pull=newer will allow us to keep this image uptodate for
users without erroring out if podman cannot check or pull a new image
(i.e. we'd keep running the local one)

### DIFF
--- a/scripts/pattern-util.sh
+++ b/scripts/pattern-util.sh
@@ -27,7 +27,7 @@ fi
 
 # Do not quote the ${KUBECONF_ENV} below, otherwise we will pass '' to podman
 # which will be confused
-podman run -it --rm \
+podman run -it --rm --pull=newer \
 	--security-opt label=disable \
 	-e EXTRA_HELM_OPTS \
 	-e KUBECONFIG \


mbaldessari:
As Martin pointed out we might be requiring a too new version of podman. So we'll need to approach this a bit differently

mbaldessari:
According to https://www.github.com/containers/podman/issues/12707 the pull newer has been present since 3.4.x days almost two years ago, so I think it is okay to start merging this one.

mbaldessari:
Tested on Ubuntu LTS as well:
```
root@ubuntu2204:~# source /etc/os-release && echo $VERSION
22.04.3 LTS (Jammy Jellyfish)
root@ubuntu2204:~# podman version
Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.18.1
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
root@ubuntu2204:~# podman run -it --rm --pull=newer quay.io/hybridcloudpatterns/utility-container:latest

       1          Welcome to the Validate Pattern utility container
      111         It contains all the needed components to install a pattern.
     1   1
    111 111       Please run the following for more information:
   1       1
  111     111       ./pattern.sh make help
 1   1   1   1
111 111 111 111
```

So merging

mhjacks:
Confirmed this works on Ubuntu LTS 22.04; this seems to cover most of our use cases (in addition to our RHEL/Fedora ones) so I'm fine with this being merged now

PR #265, opened 2023-02-02
Better Help message

mbaldessari:
- Better help output
- Drop operator-upgrade it's the same as operator-deploy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,6 @@ ansible-unittest: ## run ansible unit tests
 	pytest -r a --fulltrace --color yes ansible/tests/unit/test_*.py
 
 .PHONY: deploy upgrade legacy-deploy legacy-upgrade
-deploy upgrade legacy-deploy legacy-upgrade: ## does nothing anymore. use operator-deploy
+deploy upgrade legacy-deploy legacy-upgrade:
 	@echo "UNSUPPORTED TARGET: please switch to 'operator-deploy'"; exit 1
 
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,35 @@ TARGET_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
 
 # --set values always take precedence over the contents of -f
 HELM_OPTS=-f values-global.yaml --set main.git.repoURL="$(TARGET_REPO)" --set main.git.revision=$(TARGET_BRANCH) $(TARGET_SITE_OPT)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,16 +12,36 @@ TARGET_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
 
 # --set values always take precedence over the contents of -f
 HELM_OPTS=-f values-global.yaml --set main.git.repoURL="$(TARGET_REPO)" --set main.git.revision=$(TARGET_BRANCH) $(TARGET_SITE_OPT)
+
+##@ Pattern Common Tasks
+
 .PHONY: help
 help: ## This help message
-	@printf "$$(grep -hE '^\S.*:.*##' $(MAKEFILE_LIST) | sed -e 's/:.*##\s*/:/' -e 's/^\(.\+\):\(.*\)/\\x1b[36m\1\\x1b[m:\2/' | column -c2 -t -s :)\n"
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-16s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
mhjacks, line 20: The new parser still drops multiple-word arguments before the colon. This, for example:

```
install upgrade deploy: operator-deploy post-install ## Install or upgrade the pattern via the operator
	echo "Installed/Upgraded"
```

Which is definitely one I think we should show. :)

mhjacks (changes requested):

Makefile
mbaldessari: I hope I addressed this in my latest pass. I think we should start dropping having different targets for the same thing though, at least when we can and are sure they are not being used

mbaldessari: Reason is that when a user sees:
```
Pattern Common Tasks
  help                                 This help message
  show                                 show the starting template without installing it
  operator-deploy operator-upgrade     runs helm install
```
It is not immediately obvious that he can run both `operator-deploy` and `operator-upgrade`?

mhjacks:
Looks good. I think excluding deprecated options is fine

mhjacks (approved):

PR #270, opened 2023-02-09
Bail out on unsupported podman versions

mbaldessari:
Tested on Fedora and on Mac OSX. We use the grep -E on the '\b' word
boundary to be a bit more resilient in our grep.


claudiol (approved):
That works.  I will merge

PR #271, opened 2023-02-18
managed cluster group fix

claudiol:
- - Change to generate the RHACM PlacementRule with the correct label described in values-hub.yaml

### DIFF
--- a/.github/workflows/ansible-unittest.yml
+++ b/.github/workflows/ansible-unittest.yml
@@ -20,7 +20,7 @@ jobs:
     name: Ansible unit tests
     strategy:
       matrix:
-        python-version: [3.10.9]
+        python-version: [3.10.10]
     # Set the agent to run on
     runs-on: ubuntu-latest
 
--- a/.github/workflows/jsonschema.yaml
+++ b/.github/workflows/jsonschema.yaml
@@ -20,7 +20,7 @@ jobs:
     name: Json Schema tests
     strategy:
       matrix:
-        python-version: [3.11.1]
+        python-version: [3.11.2]
     # Set the agent to run on
     runs-on: ubuntu-latest
 
--- a/acm/templates/policies/application-policies.yaml
+++ b/acm/templates/policies/application-policies.yaml
@@ -133,7 +133,9 @@ spec:
   {{- else if (not $group.acmlabels) }}
   clusterSelector:
     matchLabels:
-      clusterGroup: {{ $group.name }}
+    {{- range $group.labels }}
+      {{ .name }}: {{ .value }}
+    {{- end }}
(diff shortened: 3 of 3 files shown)


claudiol:
@mbaldessari 
Summary: The RHACM PlacementPolicy generated for selecting clusters was using the name of the managedClusterGroup and not the label in values-hub.yaml

Details:
In values-hub.yaml we have the following definition for a **managedClusterGroup**:
```
  managedClusterGroups:
    exampleRegion:
      name: group-one
      labels:
      - name: clusterGroup
        value: group-one
      helmOverrides:
      - name: clusterGroup.isHubCluster
        value: false
```
The reason this appears to work is because we define the name of the clustergroup as *group-one* and the label as *group-one* as well. I say it appears to work because if I change the label to *foo-group* what is being generated by the ACM chart is the following:
```
# Source: acm/templates/policies/application-policies.yaml
apiVersion: apps.open-cluster-management.io/v1
kind: PlacementRule
metadata:
  name: group-one-placement
spec:
  clusterConditions:
    - status: 'True'
      type: ManagedClusterConditionAvailable
  clusterSelector:
    matchLabels:
      clusterGroup: group-one
```
Notice that the **matchLabels:** value is not correct. It should be *foo-one*.  The change in applications-policies.yaml will render the correct PlacementRule and will also handle the list of labels correctly.

Let me know if you want to chat about this change.

mbaldessari:
Shouldn't we just change the values-hub.yaml to use acmlabesl:? Otherwise we're making acmlabels and labels the exact same thing and that just feels wrong?

claudiol:
Ok I will test with acmlabels and see. But are we going to support a list of labels? As it stands right now we do not support the list of labels. So either we say it's a list of **acmlabels**, and remove the support of list of **labels**, or we say we support both list of **labels**, and **acmlabels**.

beekhof:
> Shouldn't we just change the values-hub.yaml to use acmlabesl:? Otherwise we're making acmlabels and labels the exact same thing and that just feels wrong?

IIRC, we invented labels and then very soon after added support for hosted Argo... at which point I s/labels/acmlabels/ to avoid people thinking it was relevant to Argo.  I was under the impression patterns weren't using labels when I made that change 

+1 to change the values-hub.yaml

mbaldessari:
> > Shouldn't we just change the values-hub.yaml to use acmlabesl:? Otherwise we're making acmlabels and labels the exact same thing and that just feels wrong?
> 
> IIRC, we invented labels and then very soon after added support for hosted Argo... at which point I s/labels/acmlabels/ to avoid people thinking it was relevant to Argo. I was under the impression patterns weren't using labels when I made that change
> 
> +1 to change the values-hub.yaml

Ack, that is what we ended up doing. Closing this one

PR #277, opened 2023-03-09
Add support for ~/.config/hybrid-cloud-patterns folder to store secrets

mbaldessari:
One point of feedback received lately, was that having
~/values-secrets*yaml files straight in the ~ folder can clutter the
home folder. Add support for storing them inside the ~/.patterns folder.

Overriding the location completely via VALUES_SECRET is still supported.

### DIFF
--- a/ansible/roles/vault_utils/tasks/push_secrets.yaml
+++ b/ansible/roles/vault_utils/tasks/push_secrets.yaml
@@ -63,6 +63,7 @@
     found_file: "{{ lookup('ansible.builtin.first_found', findme) }}"
   vars:
     findme:
+      - "~/.config/hybrid-cloud-patterns/values-secret-{{ pattern_name }}.yaml"
       - "~/values-secret-{{ pattern_name }}.yaml"
       - "~/values-secret.yaml"
       - "{{ pattern_dir }}/values-secret.yaml.template"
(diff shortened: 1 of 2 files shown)


ansible/roles/vault_utils/tasks/push_secrets.yaml
```diff
@@ -63,6 +63,7 @@
     found_file: "{{ lookup('ansible.builtin.first_found', findme) }}"
   vars:
     findme:
+      - "~/.patterns/values-secret-{{ pattern_name }}.yaml"
       - "~/values-secret-{{ pattern_name }}.yaml"
       - "~/values-secret.yaml"
       - "{{ pattern_dir }}/values-secret.yaml.template"
```
beekhof, line 69: At some point I would like to see this list pruned back to:
~/.patterns/values-secret-{{ pattern_name }}.yaml
{{ pattern_dir }}/values-secret.yaml.template
and VALUES_SECRET

beekhof:
lgtm

ansible/roles/vault_utils/tasks/push_secrets.yaml
mbaldessari: agree

PR #283, opened 2023-03-21
Add labels and annotations support for namespaces (RFE)

claudiol:
This PR is an initial implementation of the [Common RFE](https://github.com/hybrid-cloud-patterns/common/issues/282) created to support the addition of labels and annotations to a namespace definition.

- Added labels and annotations support to namespaces.yaml template.

This current implementation supports the following formats for namespace definitions:

```
  namespaces:
  - open-cluster-management:
      labels:
      - name: "openshift.io/node-selector"
        value: ""
      annotations:
      - name: "openshift.io/cluster-monitoring"
        value: "true"
  - application-ci
  ```

We chose to use name/value pairs for annotations and labels as an initial implementation.  Another option to consider is to use the following format:

```
  - open-cluster-management:
      labels:
        openshift.io/node-selector: ""
      annotations:
        openshift.io/cluster-monitoring: "true"
  - application-ci
```

OUTPUT:
```
---
# Source: pattern-clustergroup/templates/core/namespaces.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: open-cluster-management
  labels:
    argocd.argoproj.io/managed-by: common-example
    openshift.io/node-selector: ""
  annotations:
    openshift.io/cluster-monitoring: "true"
spec:
---
# Source: pattern-clustergroup/templates/core/namespaces.yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    argocd.argoproj.io/managed-by: common-example
  name: application-ci
spec:
---
```
We need to discuss this as a team and figure out the path forward.  

Contributors:
@mbaldessari and @claudiol

### CODE REVIEW

**Comment by beekhof** [comment]
Sorry, but I hate this.
It's backwards incompatible and makes our wrapper just as complex as the thing it creates, adding zero value.

**Comment by beekhof** [comment]
I take back the incompatibility claim, that's technically neat, but we'd no longer be adding any value over having people create the manifests directly.

**Comment by mbaldessari** [comment]
Chatted with Andrew: Maybe this is a bit premature to add and since it can make code and values files rather convoluted, maybe the first pass would be to see if adding the namespaces as templates charts is not the simpler way forward here. Can we test that and make sure that works without surprises first?

**Comment by claudiol** [comment]
Ok. Thanks for the feedback @beekhof.  I don't agree that we would not be adding value any more.  But I do agree that this seems a bit convoluted.  Going forward if a user needs to add labels to their namespaces then they would:
* Remove the namespace from the list in our values file.
* Create their own **namespace** manifest with the appropriate labels and annotations.
* Create their own **OperatorGroup** manifest

**Comment by day0hero** [comment]
imo there is value with this - 

if we ever need to override an scc for some reason we would need to be able to annotate the namespace, same idea with service mesh and things like cert-manager. The metadata.{labels,annotations} is part of the spec so it doesn't cost anything to do this.

we take a similar approach with the `env` section in subscriptions and also with the ignoreDifferences for the argo applications.

**Comment by claudiol** [comment]
Choosing second option for the namespace labels and annotations.

File: clustergroup/templates/core/namespaces.yaml
Comment (day0hero) [inline clustergroup/templates/core/namespaces.yaml:1]: /LGTM

File: tests/clustergroup-normal.expected.yaml [hunk 1]
```diff
@@ -959,11 +970,11 @@ spec:
...
 kind: OperatorGroup
 metadata:
-  name: open-cluster-management-operator-group
-  namespace: open-cluster-management
+  name: map[open-cluster-management:map[annotations:map[openshift.io/cluster-monitoring:true owner:namespace owner] labels:map[kubernetes.io/os:linux openshift.io/node-selector:]]]-operator-group
+  namespace: map[open-cluster-management:map[annotations:map[openshift.io/cluster-monitoring:true owner:namespace owner] labels:map[kubernetes.io/os:linux openshift.io/node-selector:]]]
 spec:
   targetNamespaces:
-  - open-cluster-management
+  - map[open-cluster-management:map[annotations:map[openshift.io/cluster-monitoring:true owner:namespace owner] labels:map[kubernetes.io/os:linux openshift.io/node-selector:]]]
```
beekhof, line 977: Looks broken

claudiol (commented):
Review and updated operatorgroup template.

beekhof:
/lgtm from a test perspective.
good to merge @mbaldessari ?

PR #284, opened 2023-03-28
Drop dollar sign from default password policy

mbaldessari:
Sometimes they cause issues with some usage in patterns. Let's just drop
them from the default policy.


day0hero (approved):
Definitely helps with medical-diagnosis!

PR #285, opened 2023-03-31
Bug: Fix for extraValueFiles in Application template

claudiol:
When you define the xyz chart application in the **applications:** section of the values-hub.yaml like so:
```
xyz:
      name: xyz-chart
      namespace: xyz
      project: hub
      path: charts/mgt-hub/xyz
      extraValueFiles:
      - values-myxyz.yaml
```
What we generate is the following Application manifest for ArgoCD to consume:
```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: xyz-chart
  namespace: common-example
  finalizers:
  - resources-finalizer.argocd.argoproj.io/foreground
spec:
  destination:
    name: in-cluster
    namespace: xyz
  project: hub
  source:
    repoURL: 
    targetRevision: main
    path: charts/mgt-hub/xyz
    helm:
      ignoreMissingValueFiles: true
      valueFiles:
      - "/values-global.yaml"
      - "/values-example.yaml"
      - "values-myxyz.yaml"  <============ Is this supposed to be /values-myxyz.yaml?
      # Watch the progress of https://issues.redhat.com/browse/GITOPS-891 and update accordingly
      parameters:
        - name: global.repoURL
          value: $ARGOCD_APP_SOURCE_REPO_URL
...
  syncPolicy:
    automated: {}
    #  selfHeal: true
```

This fix will generate the valuesFiles section with the '/' at the beginning of the values file:

```
      valueFiles:
      - "/values-global.yaml"
      - "/values-example.yaml"
      - "/values-myxyz.yaml"  <============ Generated as /values-myxyz.yaml
```

### DIFF
--- a/clustergroup/templates/plumbing/applications.yaml
+++ b/clustergroup/templates/plumbing/applications.yaml
@@ -60,7 +60,7 @@ spec:
             - "/values-{{ $.Values.global.clusterVersion }}-{{ $.Values.clusterGroup.name }}.yaml"
             {{- end }}
         {{- range .extraValueFiles }}
-            - {{ . | quote }}
+            - /{{ . | quote }}
         {{- end }}
         {{- if .useGeneratorValues }}
           values: |-
@@ -162,7 +162,7 @@ spec:
       - "/values-{{ $.Values.global.clusterVersion }}-{{ $.Values.clusterGroup.name }}.yaml"
       {{- end }}
       {{- range $valueFile := .extraValueFiles }}
-      - {{ $valueFile | quote }}
+      - /{{ $valueFile | quote }}
       {{- end }}
       # Watch the progress of https://issues.redhat.com/browse/GITOPS-891 and update accordingly
       parameters:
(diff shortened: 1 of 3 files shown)


mbaldessari:
I would not do this. If you put valuesExtraFile: foo.yaml, that is what you will get. If you put "/foo.yaml" that is what you'll get. With this change, if I don't know about this now I will get "//foo.yaml" which might or might not work. I think this violates the principle of least surprise.

claudiol:
@mbaldessari so why does the "/values-global.yaml" have the slash at the beginning? Where is the values file getting read from is we do not use a '/'?

claudiol:
Helm values files for overriding values in the helm chart. The path is relative to the spec.source.path directory defined.  In the example definition of an application:
  source:
    repoURL: 
    targetRevision: main
    path: charts/mgt-hub/xyz
The file will be searched at the chart/mgt-hub/xyz directory.

Closing since it's not a bug.

PR #287, opened 2023-04-03
Add an experimental letsencypt chart

mbaldessari:
This change adds an experimental letsencrypt chart that allows a pattern
user/developer to have all routes and the API endpoint use signed
certificates by letsencrypt.

At this stage only AWS is supported. The full documentation is contained
in the chart's README.md file


mhjacks (approved):
This looks entirely sensible to me. LGTM/approved

PR #288, opened 2023-04-06
gitops 1.8

mbaldessari:
- Update the gitops patterns CRD
- Switch to gitops-1.8 channel in common


mbaldessari:
Tested this on MCG, all green

PR #290, opened 2023-04-07
Move to new spec format for dex/sso

mbaldessari:
Via https://issues.redhat.com/browse/GITOPS-2761 we are told that the
dex configuration has a new format.
Old format:

    spec:
      dex:
        openShiftOAuth: true
        resources:
        ...

New format:

    spec:
      sso:
        provider: dex
        dex:
          openShiftOAuth: true
          resources:
          ...

This format is only supported starting with gitops-1.8.0, so we should
merge this only when we are absolutely sure that no pattern in no
situation needs an older gitops version.

Tested on MCG with gitops-1.8.2

Note: with this change gitops < 1.8 is not supported. Starting with
gitops-1.9 the old format will be unsupported.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -3,6 +3,7 @@
 ## May 22, 2023
 
 * Upgraded ESO to 0.8.2
+* *Important* we now use the newly blessed sso config for argo. This means that gitops < 1.8 are *unsupported*
 
 ## May 18, 2023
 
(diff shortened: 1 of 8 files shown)


mbaldessari:
Let's merge this in a few weeks, when we're reasonably sure that gitops versions < 1.8 won't be needed at all.

PR #298, opened 2023-04-20
Added health check for pvc resource in argocd.yaml

day0hero:
This allows argo to continue rolling out the rest of the applications whenever a PVC is in a pending status, which presents a chicken/egg scenario for a deployment/deploymentConfig depending on that PVC.  The storageclass bindingmode is `WaitForFirstConsumer` in the hyperscalers, and to override this we would either need to make a separate storageclass with `Immediate` volumebinding and change the order of deployment.

With health check implemented, when the pvc is pending, bound it is considered healthy, and when it reports an error will show as unhealthy. 

Without the health check the application is stuck in a progressing state and will not continue thus preventing any downstream application from deploying.

This directly affects devsecops because we create a pvc for when the pipeline runs, which is `n` minutes/hours/days after deployment. However, all other patterns are affected that create a PVC and Deployment/DeploymentConfig in separate manifests.

### DIFF
--- a/clustergroup/templates/plumbing/argocd.yaml
+++ b/clustergroup/templates/plumbing/argocd.yaml
@@ -24,8 +24,7 @@ spec:
               hs.status = "Healthy"
               hs.message = obj.status.phase
               return hs
-            end
-            if obj.status.phase == "Bound" then
+            elseif obj.status.phase == "Bound" then
               hs.status = "Healthy"
               hs.message = obj.status.phase
               return hs
--- a/clustergroup/templates/plumbing/argocd.yaml
+++ b/clustergroup/templates/plumbing/argocd.yaml
@@ -12,6 +12,8 @@ metadata:
   annotations:
     argocd.argoproj.io/compare-options: IgnoreExtraneous
 spec:
+# Adding health checks to argocd to prevent pvc resources
+# that aren't bound state from blocking deployments
   resourceCustomizations: |
     PersistentVolumeClaim:
       health.lua: |
--- a/clustergroup/templates/plumbing/argocd.yaml
+++ b/clustergroup/templates/plumbing/argocd.yaml
@@ -12,6 +12,27 @@ metadata:
   annotations:
     argocd.argoproj.io/compare-options: IgnoreExtraneous
 spec:
+  resourceCustomizations: |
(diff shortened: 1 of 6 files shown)


mbaldessari:
Is it possible to limit this healthcheck override only when the storageclass is of a certain type? I.e. when the storageclass bindingmode is WaitForFirstConsumer in this case? Or do we not get that info in lua in the plugin?

If not maybe we should consider hiding this code behind an if and a helm variable or are we sure that we absolutely always want to do this all the time? (I genuinely have no idea about potentially unintended consequences, hence me being somewhat cautious)

mbaldessari:
Also add a one-two line comment above the plugin so our future selves will remember why that thing is there ;)

claudiol:
@day0hero I would make a slight modification in the health.lua code snippet:
```
health.lua: |
        hs = {}
        if obj.status ~= nil then
          if obj.status.phase ~= nil then
            if obj.status.phase == "Pending" then
              hs.status = "Healthy"
              hs.message = obj.status.phase
              return hs
            elseif obj.status.phase == "Bound" then
              hs.status = "Healthy"
              hs.message = obj.status.phase
              return hs
            end
          end
        end
```
Since you are checking the same variable the elseif allows you to execute one block of code if the specified condition is evaluates to "Pending" and another block of code if it is evaluates to "Bound".  It feels more efficient.

day0hero:
> @day0hero I would make a slight modification in the health.lua code snippet:
> 
> ```
> health.lua: |
>         hs = {}
>         if obj.status ~= nil then
>           if obj.status.phase ~= nil then
>             if obj.status.phase == "Pending" then
>               hs.status = "Healthy"
>               hs.message = obj.status.phase
>               return hs
>             elseif obj.status.phase == "Bound" then
>               hs.status = "Healthy"
>               hs.message = obj.status.phase
>               return hs
>             end
>           end
>         end
> ```
> 
> Since you are checking the same variable the elseif allows you to execute one block of code if the specified condition is evaluates to "Pending" and another block of code if it is evaluates to "Bound". It feels more efficient.

Sure, I'll make the change and test.

day0hero:
Tested the `elseif` change and it worked fine.

mbaldessari:
Okay let's merge this then!

PR #303, opened 2023-05-15
Allow overriding the pattern's name

mbaldessari:
This is especially useful when multiple people are working on a pattern
an have been using different names:

    $ make help |grep Pattern:
    Pattern: multicloud-gitops
    $ make NAME=foobar help |grep Pattern:
    Pattern: foobar


mhjacks:
lgtm

PR #322, opened 2023-07-06
Expanded subchart folders

mbaldessari:
- Switch to using ./subcharts as the unpacked folder for subcharts that need patching
- Do not consider subcharts/ when testing charts
- Final commit to switch to exploded subchart folders
- Drop markdown from super-linter
- Switch to using ./subcharts as the unpacked folder for subcharts that need patching (hashicorp-vault)
- Final commit to switch to exploded subchart folders


claudiol (approved):
Thanks for doing this @mbaldessari.  Everything here is in line with what was tested.  Approved.

mbaldessari:
Let's first solve the retries breakage and then we retackle this one (https://github.com/hybrid-cloud-patterns/common/pull/323)

mbaldessari:
So in my tests, I did not need this at all? I.e. I hit no errors and it seems to me that the packages subcharts were picked up just fine. Closing for now, we can reassess if needed

PR #323, opened 2023-07-07
Revert ESO/Vault Job and add a default higher number of retries

mbaldessari:
- Revert "Make Test"
- Revert "Adding rbac to support the vault sa checking on the vault-0 pod status."
- Revert "Fix image and comment"
- Revert "Add tests"
- Revert "WIP add presync for eso that waits for vault to be up"
- Increase the default retry limit when syncing


claudiol:
@mbaldessari If you look at the ArgoCD documentation for [applications.yaml](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/application.yaml) you can see that we can set the retry to a negative number if we want it to retry unlimited times.  @day0hero is testing this.  Once tested we might want to have that be the default value i.e. -1.

day0hero:
lgtm

I successfully tested these changes on multicluster-devsecops and medical-diagnosis

mbaldessari:
Thanks guys, merging!

PR #334, opened 2023-07-28
Fix sa/namespace mixup in vault_spokes_init

stocky37:
Noticed that the service account & namespace vars were mixed around for spoke vault configs.

It is not normally noticeable since they default to the same value, but in my situation I have them different.

This led to noticing the following error in my clustersecretstores:

```
Code: 403. Errors:

* namespace not authorized`
```


mbaldessari:
Good catch, thanks @stocky37!

PR #342, opened 2023-08-15
Add extraValueFiles to operator-install chart

mhjacks:


mbaldessari:
I think you also want:
```
diff --git a/clustergroup/values.schema.json b/clustergroup/values.schema.json
index e64a812..b1359a3 100644
--- a/clustergroup/values.schema.json
+++ b/clustergroup/values.schema.json
@@ -83,6 +83,10 @@
             }
           }
         },
+        "extraValueFiles": {
+          "type": "array",
+          "description": "List of extra values files that will be passed to ArgoCD."
+        },
         "gitops": {
           "type": "object",
           "additionalProperties": false,
```

Or something along those lines anyways

mhjacks:
Do we need to talk about whether those extraValueFiles are "main" or "global"? It seems like they should be "global" since I think we intend them to also apply to subsequent applications?

mbaldessari:
Yeah so I never really liked that main vs global distinction. In theory as far as I understood things 'main' is for things passed at 'make install' time and global is everything else. I think the distinction is non obvious so I would not hate the idea of just moving towards global

PR #364, opened 2023-09-09
Rework installation target

mbaldessari:
Drop the old logic and just install the CRD via OC and use helm template
for the rest.

The rationale is that helm install is very picky whenever it encounters
things that already exist. We have three potential scenarios at work
here:
A) User installs operator+pattern via CLI and updates the pattern via
   CLI (this worked before this change as well)
B) User installs operator+pattern via UI but runs updates (changing
   branch for example) via CLI (this worked before this change as well)
C) User installs only the operator via UI. Installs and updated the
   pattern via CLI. This was broken before this change. The error you'd
   get was:
   ```
   ./pattern.sh make install
   ...
   https://github.com/mbaldessari/multicloud-gitops.git - branch main: Running inside a container: Skipping git ssh checks
   + oc get crds patterns.gitops.hybrid-cloud-patterns.io
   + echo 'Reapplying helm chart:'
   Reapplying helm chart:
   + helm template --name-template multicloud-gitops common/operator-install/ -f values-global.yaml --set main.git.repoURL=https://github.com/mbaldessari/multicloud-gitops.git --set main.git.revision=main
   + oc apply set-last-applied --create-annotation -f-
   WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/michele/sno1-kubeconfig
   Error from server (NotFound): patterns.gitops.hybrid-cloud-patterns.io "multicloud-gitops" not found
   ```

With this change we simplify the process and we forcefully apply/install
the CRD for patterns via the oc command. And then we simply template out
the operator-install chart and oc apply it. We retry it a few times,
because the CRD might not yet be fully registered in the cluster.

Tested with on the A), B) and C) scenarios successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,21 +44,15 @@ help: ## This help message
 show: ## show the starting template without installing it
 	helm template common/operator-install/ --name-template $(NAME) $(HELM_OPTS)
 
-# Only call helm install if the CRD is missing. If it already exists just
-# push the templated files.
-# The reason we have two helm template calls in the else branch is to avoid
-# warnings when the chart gets applied the first time, but the resources were
-# created first via the VP operator's UI
 .PHONY: operator-deploy
 operator-deploy operator-upgrade: validate-prereq validate-origin ## runs helm install
-	@set -e; if ! oc get crds patterns.gitops.hybrid-cloud-patterns.io >/dev/null 2>&1; then \
-	  echo "Running helm:"; \
-	  helm upgrade --install $(NAME) common/operator-install/ $(HELM_OPTS); \
-	else \
-	  echo "Reapplying helm chart:"; \
-	  helm template --name-template $(NAME) common/operator-install/ $(HELM_OPTS) | oc apply set-last-applied --create-annotation -f-; \
-	  helm template --name-template $(NAME) common/operator-install/ $(HELM_OPTS) | oc apply -f-; \
-	fi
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -44,21 +44,15 @@ help: ## This help message
...
-	  helm template --name-template $(NAME) common/operator-install/ $(HELM_OPTS) | oc apply set-last-applied --create-annotation -f-; \
-	  helm template --name-template $(NAME) common/operator-install/ $(HELM_OPTS) | oc apply -f-; \
-	fi
+	@set -e -o pipefail
+	# We reapply the CRD and if it already exists we do not error out
+	oc apply common/operator-install/crds/*.yaml 2>/dev/null || /bin/true
+	# Retry five times because the CRD might not be fully installed yet
+	for i in {1..5}; do \
+		helm template --name-template $(NAME) common/operator-install/ $(HELM_OPTS) | oc apply -f- && break || sleep 10; \
+	done
```
beekhof, line 55: I think you want `helm template --include-crds` instead of this loop

mbaldessari: Nah you cannot avoid that loop :/ (believe me, I tried ;) ) . oc apply is not smart enough to wait for the CRD to be registered in the cluster. What I can do is drop the initial oc apply call for the CRD and just add --include-crds in the loop

PR #366, opened 2023-09-14
add awstools into common

day0hero:
Added aws management targets to Makefile and updated the pattern-utils.sh script to mount the aws credentials inside the container.


day0hero:
Cancelling PR since it is out of scope for common repo.

PR #374, opened 2023-10-02
Template values

beekhof:


mbaldessari:
Tested both the vanilla case and an mcg with values-hub tweaked to use both sharedValueFiles and extraValueFiles

Issue #296, opened 2023-04-14
override/empty  target namespace 


Ladanow:
Some operators are not compatible with adding targetNamespace as their own, this is automatically done when we create a new namespace.

for example, `performance-addon-operator` can't monitor its own namespace.

```
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operators.coreos.com/v1","kind":"OperatorGroup","metadata":{"annotations":{},"labels":{"app.kubernetes.io/instance":"npss-tnc-hub"},"name":"openshift-performance-addon-operator-operator-group","namespace":"openshift-performance-addon-operator"},"spec":{"targetNamespaces":["openshift-performance-addon-operator"]}}
```

claudiol:
We might want to add a way to control the parameters that VP uses while creating operatorgroups for operator installation.  The framework always creates an OperatorGroup with spec.targetNamespaces and some operators want an OperatorGroup without targetNamespaces.
@mbaldessari

PR #378, opened 2023-10-13
Adding key to exclude target namespace in operatorgroup

claudiol:
Implements the issue describe in [Issue #296 ](https://github.com/validatedpatterns/common/issues/296)

- Added operatorgroupExcludeTargetNS key to exclude targetNamespace in an Operator Group

### DIFF
--- a/clustergroup/templates/core/operatorgroup.yaml
+++ b/clustergroup/templates/core/operatorgroup.yaml
@@ -12,7 +12,7 @@ metadata:
   name: {{ $k }}-operator-group
   namespace: {{ $k }}
 spec:
-    {{- if not ( has $k $.Values.clusterGroup.operatorgroupExcludeTargetNS ) }}
+    {{- if not $v.excludeOperatorGroupTargetNS }}
   targetNamespaces:
   - {{ $k }}
     {{- end }}
@@ -24,10 +24,8 @@ metadata:
   name: {{ . }}-operator-group
   namespace: {{ . }}
 spec:
-    {{- if not ( has . $.Values.clusterGroup.operatorgroupExcludeTargetNS ) }}
   targetNamespaces:
   - {{ . }}
-    {{- end }}
   {{- end }} {{- /* if kindIs "string" $ns */}}
 ---
 {{- end }} {{- /* if or (empty $.Values.clusterGroup.operatorgroupExcludes) (not (has . $.Values.clusterGroup.operatorgroupExcludes)) */}}
--- a/clustergroup/templates/core/operatorgroup.yaml
+++ b/clustergroup/templates/core/operatorgroup.yaml
@@ -12,10 +12,14 @@ metadata:
   name: {{ $k }}-operator-group
   namespace: {{ $k }}
 spec:
-    {{- if not $v.excludeOperatorGroupTargetNS }}
   targetNamespaces:
+    {{- if $v.operatorgroup }}
+      {{- range $v.operatorgroup.targetNamespaces }}{{- /* We loop through the list of tergetnamespaces */}}
+  - {{ . }}
+      {{- end }}{{- /* End range targetNamespaces */}}
+    {{- else }}
   - {{ $k }}
-    {{- end }}
+    {{- end }}{{- /* End of if operatorGroup */}}
   {{- end }}{{- /* range $k, $v := $ns */}}
   {{- else if kindIs "string" $ns }}
 apiVersion: operators.coreos.com/v1
--- a/clustergroup/templates/core/operatorgroup.yaml
+++ b/clustergroup/templates/core/operatorgroup.yaml
@@ -15,7 +15,6 @@ spec:
   targetNamespaces:
   - {{ $k }}
   {{- end }}{{- /* range $k, $v := $ns */}}
-
   {{- else if kindIs "string" $ns }}
 apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
@@ -23,8 +22,10 @@ metadata:
   name: {{ . }}-operator-group
   namespace: {{ . }}
 spec:
+    {{- if not ( has . $.Values.clusterGroup.operatorgroupExcludeTargetNS ) }}
   targetNamespaces:
   - {{ . }}
+    {{- end }}
   {{- end }} {{- /* if kindIs "string" $ns */}}
 ---
 {{- end }} {{- /* if or (empty $.Values.clusterGroup.operatorgroupExcludes) (not (has . $.Values.clusterGroup.operatorgroupExcludes)) */}}
--- a/clustergroup/templates/core/operatorgroup.yaml
+++ b/clustergroup/templates/core/operatorgroup.yaml
@@ -12,8 +12,10 @@ metadata:
   name: {{ $k }}-operator-group
   namespace: {{ $k }}
 spec:
+    {{- if not ( has $k $.Values.clusterGroup.operatorgroupExcludeTargetNS ) }}
   targetNamespaces:
   - {{ $k }}
+    {{- end }}
   {{- end }}{{- /* range $k, $v := $ns */}}
   {{- else if kindIs "string" $ns }}
 apiVersion: operators.coreos.com/v1
--- a/clustergroup/templates/core/operatorgroup.yaml
+++ b/clustergroup/templates/core/operatorgroup.yaml
@@ -13,7 +13,7 @@ metadata:
   namespace: {{ $k }}
 spec:
   targetNamespaces:
-    {{- if or (and $v.operatorGroup (hasKey $v "targetNamespaces")) (and $v.operatorGroup (not (empty $v.targetNamespaces))) }}
+    {{- if (hasKey $v "targetNamespaces") }}
       {{- range $v.targetNamespaces }}{{- /* We loop through the list of tergetnamespaces */}}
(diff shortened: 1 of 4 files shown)


clustergroup/values.schema.json
```diff
@@ -266,6 +266,13 @@
             "type": "string"
           }
         },
+        "operatorgroupExcludeTargetNS": {
+          "type": "array",
+          "description": "List of namespaces to exclude the target namespace.",
```
mbaldessari, line 271: Maybe call it `Specify the list of namespaces where the target namespace field in the corresponding operatorgroup object should be excluded`

mbaldessari (commented):
Can you also add a small example in examples/values-example.yaml?

I am not entirely sure about the API/naming used. On one side there is precedent already via the operatorgroupExcludes and on the other hand I am not super fond of adding yet another variable. Maybe we should consider adding a field in the namespace list, like we do for annotations and labels? (I do not love this one either, maybe it's because it is Monday :) )

clustergroup/values.schema.json
claudiol: Done.

claudiol:
@mbaldessari Potential alternative if we want to exclude target namespace from OperatorGroup:

```
  namespaces:
  - open-cluster-management:
      labels:
        openshift.io/node-selector: ""
        kubernetes.io/os: linux
      annotations:
        openshift.io/cluster-monitoring: "true"
        owner: "namespace owner"
      excludeOperatorGroupTargetNS: true
  - application-ci:
      excludeOperatorGroupTargetNS: true
  - include-ci
  - excludes-ci

  operatorgroupExcludes:
  - excludes-ci
  ```

**Comment by claudiol** [comment]
> @mbaldessari Potential alternative if we want to exclude target namespace from OperatorGroup:
> 
> ```
>   namespaces:
>   - open-cluster-management:
>       labels:
>         openshift.io/node-selector: ""
>         kubernetes.io/os: linux
>       annotations:
>         openshift.io/cluster-monitoring: "true"
>         owner: "namespace owner"
>       excludeOperatorGroupTargetNS: true
>   - application-ci:
>       excludeOperatorGroupTargetNS: true
>   - include-ci
>   - excludes-ci
> 
>   operatorgroupExcludes:
>   - excludes-ci
> ```
Here's a updated approach to include operatorgroup definitions inside the namespaces construct.
- The idea is to eventually deprecate **operatorgroupExcludes**
- Add the **operatorgroup** to the definition of namespaces in our values files when defined as a map

The construct would look something like this:
1. Define the namespace as a map and explicitly exclude the **namespace** from  **targetNamespaces** in the OperatorGroup manifest
```
   namespaces:
   - application-ci:
     labels:
        openshift.io/node-selector: ""
        kubernetes.io/os: linux
     annotations:
        openshift.io/cluster-monitoring: "true"
        owner: "namespace owner"
     operatorgroup:
        targetNamespaces:
```
2. Define the namespace as a map and include a list of **targetNamespaces** in the OperatorGroup manifest
```
   namespaces:
   - application-ci:
     labels:
        openshift.io/node-selector: ""
        kubernetes.io/os: linux
     annotations:
        openshift.io/cluster-monitoring: "true"
        owner: "namespace owner"
     operatorgroup:
        targetNamespaces:
          - application-ci
          - namespace2
```
``NOTE: You must explicitly list your own namespace. It will not included by the framework in the generated manifest``
3. Define the namespace as a map and not including the **operatorgroup** key. By default this will add the **namespace** to **targetNamespaces**
```
   namespaces:
   - application-ci:
     labels:
        openshift.io/node-selector: ""
        kubernetes.io/os: linux
     annotations:
        openshift.io/cluster-monitoring: "true"
        owner: "namespace owner"
```
4. If the namespace is defined as a list the default is to include the **namespace** in  **targetNamespaces**
``` 
  namespaces:
  - application-ci
```
5. To prevent the framework to create an OperatorGroup for the namespace we will still include the namespace in the **operatorgroupExcludes** list.

File: examples/values-example.yaml [hunk 2]
```diff
@@ -23,13 +23,20 @@ clusterGroup:
       annotations:
         openshift.io/cluster-monitoring: "true"
         owner: "namespace owner"
-  - application-ci
+  - application-ci:
+      operatorgroup:
+        targetNamespaces:
+          - application-ci
+          - other-namespace
```
beekhof, line 30: I would suggest decoupling `operatorgroup` and `targetNamespaces`
Also, be consistent with camelCase.

```suggestion
      operatorGroup: true # true being the default anyway
      targetNamespaces:
        - application-ci
        - other-namespace
```

examples/values-example.yaml
```diff
@@ -23,13 +23,20 @@ clusterGroup:
...
+  - exclude-targetns:
+      operatorgroup:
+        targetNamespaces:
+  - include-ci
   - excludes-ci
```
beekhof, line 35:
```suggestion
  - excludes-ci
     operatorGroup: false
```

examples/values-example.yaml
claudiol: Thanks for the suggestion and it totally makes sense.  Updated code to support this.

examples/values-example.yaml
claudiol: Updated values-example.yaml with change.

claudiol:
> > @mbaldessari Potential alternative if we want to exclude target namespace from OperatorGroup:
> > ```
> >   namespaces:
> >   - open-cluster-management:
> >       labels:
> >         openshift.io/node-selector: ""
> >         kubernetes.io/os: linux
> >       annotations:
> >         openshift.io/cluster-monitoring: "true"
> >         owner: "namespace owner"
> >       excludeOperatorGroupTargetNS: true
> >   - application-ci:
> >       excludeOperatorGroupTargetNS: true
> >   - include-ci
> >   - excludes-ci
> > 
> >   operatorgroupExcludes:
> >   - excludes-ci
> > ```
> 
> Here's a updated approach to include operatorgroup definitions inside the namespaces construct.
> 
> * The idea is to eventually deprecate **operatorgroupExcludes**
> * Add the **operatorgroup** to the definition of namespaces in our values files when defined as a map
> 
> The construct would look something like this:
> 
> 1. Define the namespace as a map and explicitly exclude the **namespace** from  **targetNamespaces** in the OperatorGroup manifest
> 
> ```
>    namespaces:
>    - application-ci:
>      labels:
>         openshift.io/node-selector: ""
>         kubernetes.io/os: linux
>      annotations:
>         openshift.io/cluster-monitoring: "true"
>         owner: "namespace owner"
>      operatorgroup:
>         targetNamespaces:
> ```
> 
> 2. Define the namespace as a map and include a list of **targetNamespaces** in the OperatorGroup manifest
> 
> ```
>    namespaces:
>    - application-ci:
>      labels:
>         openshift.io/node-selector: ""
>         kubernetes.io/os: linux
>      annotations:
>         openshift.io/cluster-monitoring: "true"
>         owner: "namespace owner"
>      operatorgroup:
>         targetNamespaces:
>           - application-ci
>           - namespace2
> ```
> 
> `NOTE: You must explicitly list your own namespace. It will not included by the framework in the generated manifest` 3. Define the namespace as a map and not including the **operatorgroup** key. By default this will add the **namespace** to **targetNamespaces**
> 
> ```
>    namespaces:
>    - application-ci:
>      labels:
>         openshift.io/node-selector: ""
>         kubernetes.io/os: linux
>      annotations:
>         openshift.io/cluster-monitoring: "true"
>         owner: "namespace owner"
> ```
> 
> 4. If the namespace is defined as a list the default is to include the **namespace** in  **targetNamespaces**
> 
> ```
>   namespaces:
>   - application-ci
> ```
> 
> 5. To prevent the framework to create an OperatorGroup for the namespace we will still include the namespace in the **operatorgroupExcludes** list.

We now use the following namespace structure to include/exclude an OperatorGroup:

1. Create OperatorGroup for the Namespace and exclude targetNamespaces from the OperatorGroup definition
```
   namespaces:
   - application-ci:
     labels:
        openshift.io/node-selector: ""
        kubernetes.io/os: linux
     annotations:
        openshift.io/cluster-monitoring: "true"
        owner: "namespace owner"
     operatorGroup: true
     targetNamespaces:
```
2. Do not create the OperatorGroup for the Namespace
```
   namespaces:
   - exclude-operatorgroup:
     operatorGroup: false
```
3. Create the OperatorGroup for the Namespace and add the namespace to the targetNamespaces
```
   namespaces:
   - include-targetns:
     operatorGroup: true
```
Same result if you just have it as a list and not a map
```
   namespaces:
   - include-targetns
```

clustergroup/templates/core/operatorgroup.yaml
```diff
@@ -5,17 +5,23 @@
...
-
+  {{- if $v.operatorGroup }}{{- /* Checks if the user sets operatorGroup: false */}}
 apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
 metadata:
   name: {{ $k }}-operator-group
   namespace: {{ $k }}
 spec:
   targetNamespaces:
+    {{- if or (and $v.operatorGroup (hasKey $v "targetNamespaces")) (and $v.operatorGroup (not (empty $v.targetNamespaces))) }}
```
beekhof, line 16:
```suggestion
    {{- if (hasKey $v "targetNamespaces") }}
```

PR #380, opened 2023-10-19
Support a live preview of apps against a running cluster

beekhof:


mhjacks:
*TL;DR:* `make preview-all` for a relatively intelligent rendering of helm charts with your designated overrides

The scripts determine overrides by "knowing" what the VP operator will add, and further introspect values-global and the other values files to extract and include the designated override mechanisms.

I'd guess that it will cover >80% of OpenShift GitOps use cases in the VP framework currently.

This has some (known) signfiicant limitations, which can be addressed, but might need a different approach to do so:

* Parsing of applications needs to be more clever - currently the code assumes that all targets will be local charts. This is not true, for example, in industrial-edge.
* There is currently not a mechanism to actually preview against two different clusters (i.e. a hub and a remote). All previews will be done against the hub cluster, even though the script is pointed at a different clustergroup file.
* Make output can be included in the YAML. This can be avoided with explicit redirects etc but that would require us to be more opinionated about output than I think we should be at this point.

Some of these are going to be harder to address than others. I think that if this approach is deemed valuable, the right way to fix the deficiencies will be to write some python code (that could use the python interpreter and modules inside the utility container) to do the parsing and output.

PR #391, opened 2023-11-07
Parametrize ESO caProvider fields

mbaldessari:
This will allow users to disable using the default kube-root-ca.crt
configmap and/or customize things fully.

So, for example, an environment which uses letsencrypt for the ingress
certs, but does not have that CA included in kube-root-ca.crt, can
simply override things with a /values-IBMCloud.yaml file with:

caProvider:
  enabled: false

Which will just use the CA System bundle to authenticate connections
from ESO to vault.

### DIFF
--- a/golang-external-secrets/values.yaml
+++ b/golang-external-secrets/values.yaml
@@ -1,6 +1,22 @@
 ---
+# Eventually we should aim to move these two under the golangExternalSecrets key
 mountPath: "hub"
 mountRole: "hub-role"
+
+golangExternalSecrets:
+  # This controls how ESO connects to vault
+  caProvider:
+    enabled: true # If vault is exposed via a route that is signed by a non internal CA you might want to disable this
+    hubCluster:
+      type: ConfigMap
+      name: kube-root-ca.crt
+      key: ca.crt
+      namespace: golang-external-secrets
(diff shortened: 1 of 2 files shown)


claudiol (approved):
Very reasonable approach.  
/LGTM

mbaldessari:
> /lgtm as far as it goes - maybe with this merged we can put this into the MCGO with the explicit override mechanism?

Aye, that is the plan. So we showcase the cloud overrides a bit better.

golang-external-secrets/values.yaml
```diff
@@ -1,7 +1,23 @@
 ---
+# Eventually we should aim to move these two under the golangExternalSecrets key
 mountPath: "hub"
 mountRole: "hub-role"
 
+golangExternalSecrets:
+  # This controls how ESO connects to vault
+  caProvider:
+    enabled: true # If vault is exposed via a route that is signed by a non internal CA you might want to disable this
+    hubCluster:
```
beekhof, line 10: Can we avoid the hub/nonhub terminology please?  The world shouldn't break, and should still make sense, if someone doesn't start with a clustergroup called "hub".

What about `vaultHostCluster`, and `vaultClientCluster` which better describe the roles here?

PR #394, opened 2023-11-14
Update ConfigManagementPlugins handling

mhjacks:
1) Make configmanagement plugins optional
2) Allow for old style cmp's via legacyConfigManagementPlugins
3) All for "CMP 2.0" plugins aka sidecar plugins (now preferred)
4) Add new argoCD key to clusterGroup values with needed schema changes


mhjacks:
Note: experimentation seems to indicate that CMP 2.0 deliberately makes it difficult (that is, practically impossible) to run executables inside the repository, so this feature will not support doing that with CMP 2.0. Users who really need that can used the CMP 1.0 framework (Which this PR supports) or plan to migrate.

clustergroup/templates/plumbing/argocd-cmp-plugin-cms.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if len $.Values.clusterGroup.argoCD.configManagementPlugins }}
+{{- range $cmp := $.Values.clusterGroup.argoCD.configManagementPlugins }}
```
beekhof, line 2: Is this if clause redundant here?

examples/kustomize-renderer/kustomize
beekhof, line 1: If we don't test plugins, we also won't know if we break the templates we use to drive them.
Maybe we still need this example chart?

clustergroup/templates/plumbing/argocd-cmp-plugin-cms.yaml
mhjacks: Yes, it is. Fixed.

examples/kustomize-renderer/kustomize
mhjacks: True. I re-added it.

clustergroup/templates/plumbing/argocd.yaml
```diff
@@ -36,36 +36,9 @@ spec:
...
-            --set global.namespace=$ARGOCD_APP_NAMESPACE
-            --set global.pattern={{ .Values.global.pattern }}
-            --set global.clusterDomain={{ .Values.global.clusterDomain }}
-            --set global.hubClusterDomain={{ .Values.global.hubClusterDomain }}
-            --set global.localClusterDomain={{ coalesce .Values.global.localClusterDomain .Values.global.hubClusterDomain }}
-            --set clusterGroup.name={{ .Values.clusterGroup.name }}
-            --post-renderer ./kustomize"]
+  {{- if len $.Values.clusterGroup.argoCD.legacyConfigManagementPlugins }}
+  configManagementPlugins: | {{ $.Values.clusterGroup.argoCD.legacyConfigManagementPlugins | nindent 4 }}
+  {{- end }}
```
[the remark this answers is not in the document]
mhjacks: No, it's redundant. I coded this before I knew that the mechanism had been removed in current upstream. It's gone now.

PR #403, opened 2023-11-29
Private repo support

mbaldessari:
- Add support for private repos
- Amend tests

### DIFF
--- a/acm/templates/policies/private-repo-policies.yaml
+++ b/acm/templates/policies/private-repo-policies.yaml
@@ -1,0 +1,161 @@
+# We copy the vp-private-repo-credentials from the "openshift-gitops" namespace
+# to the "open-cluster-management" via the "private-hub-policy"
+#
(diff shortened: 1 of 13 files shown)


acm/templates/policies/private-repo-policies.yaml
```diff
@@ -0,0 +1,161 @@
+# We copy the vp-private-repo-credentials from the "openshift-gitops" namespace
+# to the "open-cluster-management" via the "private-hub-policy"
+#
+# Then we copy the secret from the "open-cluster-management" namespace to the
+# managed clusters "openshift-gitops" instance
+#
+# And we also copy the same secret to the namespaced argo's namespace
```
beekhof, line 7: Oh wow.  But I also imagine we don't have any other choice

mbaldessari: Yeah, we could drive this imperatively via ansible, but that is worse than this imo

PR #409, opened 2023-12-06
Make the unsealJob conditional to the vault application's existence

mbaldessari:
This way, if there is no vault application we just don't define this job
at all.

The problem with this approach is that it is quite opinionated on the
name of the application. So it will break if you call your app something
else than 'vault'. I kind of prefer this approach to adding yet another
variable though


claudiol:
@mbaldessari I agree that this will break if the name of the application is not vault.  As we discussed maybe using the path might be a better choice.

mbaldessari:
we'll probable address this with the new secrets backends

PR #461, opened 2024-04-11
Custom CA support

mbaldessari:
- **Add support for custom CAs**
- **Update tests**

### DIFF
--- a/acm/templates/policies/ocp-gitops-policy.yaml
+++ b/acm/templates/policies/ocp-gitops-policy.yaml
@@ -1,3 +1,4 @@
+{{- $hasInitContainerCapability := and (.Values.global.experimentalCapabilities) (has "initcontainers" (splitList "," .Values.global.experimentalCapabilities)) }}
 apiVersion: policy.open-cluster-management.io/v1
 kind: Policy
 metadata:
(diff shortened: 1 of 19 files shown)


acm/templates/policies/ocp-gitops-policy.yaml
```diff
@@ -44,6 +54,163 @@ spec:
...
+                    scopes: '[groups]'
+                  redis:
+                    resources:
+                      limits:
+                        cpu: 500m
+                        memory: 256Mi
+                      requests:
+                        cpu: 250m
+                        memory: 128Mi
+                  repo:
```
mhjacks, line 136: How will this play with the new plugins code? My assumption is "not well", which doesn't mean we can't do this but it might take a release note or something to say whether we can or can't

mhjacks (commented):
This is clearly an important capability and the changes here make sense. What I'd like to see is a test for specifying both this feature and a custom repo plugin at the same time to see what happens. If it works, great - if not, we should note that it doesn't. I suspect the usage of the features will intersect, since the main use of the plugins is to add support for PolicyGen and the things that want PolicyGen seem to be predominantly internal.

claudiol (approved):
Overall these changes are needed for customers that require their own CA applied to their OpenShift clusters.  As it's marked "experimental" we can ask one of our customers, or partners, to test in their environment and iterate through the issues encountered if any.

acm/templates/policies/ocp-gitops-policy.yaml
mhjacks: Upon further review, these features should be orthogonal.

PR #463, opened 2024-04-12
feat: added clusterdeployment list per managedclustergroup

TomerFi:


TomerFi:
Oops, wrong target.
Once done, I will link the new PR to the proper target.

PR #499, opened 2024-05-02
New global scope argocdServer section for values-global.yaml

claudiol:
- Added new section for to configure the ArgoCD server to support tls
```
global:
  argocdServer:
    route:
      tls:
        insecureEdgeTerminationPolicy: Redirect
        termination: reencrypt
```
- Default for ArgoCD is to create route with the following:
```
      route: enabled
      tls:
        insecureEdgeTerminationPolicy: Redirect
        termination: passthrough
```

    For more information please refer to https://issues.redhat.com/browse/GITOPS-3918.

### DIFF
--- a/acm/templates/policies/ocp-gitops-policy.yaml
+++ b/acm/templates/policies/ocp-gitops-policy.yaml
@@ -195,6 +195,11 @@ spec:
                         memory: 128Mi
                     route:
                       enabled: true
+                      {{- if and (.Values.global.argocdServer) (.Values.global.argocdServer.route) (.Values.global.argocdServer.route.tls) }}
+                      tls:
+                        insecureEdgeTerminationPolicy: {{ default "Redirect" .Values.global.argocdServer.route.tls.insecureEdgeTerminationPolicy }}
+                        termination: {{ default "reencrypt" .Values.global.argocdServer.route.tls.termination }}
+                      {{- end }}
                     service:
                       type: ""
                   sso:


acm/templates/policies/ocp-gitops-policy.yaml
```diff
@@ -195,6 +195,11 @@ spec:
                         memory: 128Mi
                     route:
                       enabled: true
+                      {{- if and (.Values.global.argocdServer) (.Values.global.argocdServer.route) (.Values.global.argocdServer.route.tls) }}
+                      tls:
+                        insecureEdgeTerminationPolicy: {{ default "Redirect" $.Values.global.argocdServer.route.tls.insecureEdgeTerminationPolicy }}
+                        termination: {{ default "passthrough" $.Values.global.argocdServer.route.tls.termination }}
```
mbaldessari, line 201: In the operator we went for the reencrypt default, so we should do the same here?

claudiol: @mbaldessari  Sure. Let me change that for consistency.

claudiol: @mbaldessari Pushed the changes.

mbaldessari:
We should do this also in `clustergroup/templates/argocd.yaml` and tweak the schema for clustergroup (if needed). That way we're a bit more consistent with things.

mbaldessari:
We'll fix up the inconsistency in a separate pr

PR #501, opened 2024-05-03
Drop gitopsspec from pattern's CR

mbaldessari:
We drive this from the patterns-operator-config configmap these days, which
makes more sense (it is a clusterwide setting and not really a per
pattern one).


claudiol (approved):
/lgtm

PR #505, opened 2024-05-09
Added support to enable user workloads in control plane nodes

soukron:
Simple PR to allow setting control planes schedulables without extra charts.


soukron:
Let's consider it a WIP. I'd prefer to complete the PR by allowing to control the scheduler completely to control the masters, defaultNodeSelector and a profile.

PR #507, opened 2024-05-09
Simplified PR for auto approve install plans

soukron:
Simplified version of https://github.com/validatedpatterns/common/pull/480 as nowadays the imperative service account already has cluster-admin.


mbaldessari:
Thanks! Note that there is a separate admin user for the imperative SA. I tweaked that here https://github.com/validatedpatterns/common/pull/525 (the old one has the same old permissions as usual)

PR #513, opened 2024-05-16
WIP: Additional namespace object definition for ACM Policy

claudiol:
- In some environments we have noticed that in some instances the imperative namespace does not get created on
  Spoke clusters.
- The policy in common/acm/templates/policies/acm-hub-ca-policy creates a Secret in the imperative namespace which
  might fail if the namespace does not exist.
- Adding new objectDefinition to the current ACM policy to create the 'imperative' namespace.
- Updated CI Tests with new ACM Policy

### DIFF
--- a/acm/templates/policies/acm-hub-ca-policy.yaml
+++ b/acm/templates/policies/acm-hub-ca-policy.yaml
@@ -24,6 +24,12 @@ spec:
             include:
               - default
           object-templates:
+            - complianceType: musthave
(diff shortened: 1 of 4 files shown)


mbaldessari:
I would first try and understand why the imperative namespace does not get created on the spoke, no? The clusterwide argo on the spoke is in charge of that and eventually it appears and then the acm policy will get applied

PR #518, opened 2024-05-17
Updates to acm-hub-ca

mbaldessari:
- **Use golang-external-secrets for the acm hub-ca bits**
- **Only do the acm hub ca policy when vault is the backend**


claudiol (approved):
/lgtm
Changes will help with namespace creation issues

PR #521, opened 2024-05-29
bug: Invalid OperatorGroup generated when omitting targetNamespaces

claudiol:
bug: Invalid OperatorGroup generated when ommitting targetNamespaces

Problem Statement:
When setting a namespace like this:
    - openshift-distributed-tracing:
        operatorGroup: true
        targetNamespaces: []

The chart generates the following yaml:
```yaml
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  name: openshift-distributed-tracing-operator-group
  namespace: openshift-distributed-tracing
spec:
  targetNamespaces:
```

Which k8s rejects the targetNamespaces key as invalid when it attempts to apply it and removes it since it doesn't have a value, which just so happens to have the desired result of not setting the targetNamespaces (or a selector) to enable it for All Namespaces.

### DIFF
--- a/clustergroup/templates/_helpers.tpl
+++ b/clustergroup/templates/_helpers.tpl
@@ -202,13 +202,19 @@ kind: OperatorGroup
 metadata:
   name: {{ $k }}-operator-group
   namespace: {{ $k }}
+      {{- if (hasKey $v "targetNamespaces") }}
+        {{- if $v.targetNamespaces }}
+          {{- if (len $v.targetNamespaces) }}
 spec:
   targetNamespaces:
-      {{- if (hasKey $v "targetNamespaces") }}
-        {{- range $v.targetNamespaces }}{{- /* We loop through the list of tergetnamespaces */}}
+            {{- range $v.targetNamespaces }}{{- /* We loop through the list of tergetnamespaces */}}
   - {{ . }}
-        {{- end }}{{- /* End range targetNamespaces */}}
+            {{- end }}{{- /* End range targetNamespaces */}}
+          {{- end }}{{- /* End if (len $v.targetNamespaces) */}}
+        {{- end }}{{- /* End $v.targetNamespaces */}}
       {{- else }}
+spec:
+  targetNamespaces:
   - {{ $k }}
       {{- end }}{{- /* End of if hasKey $v "targetNamespaces" */}}
     {{- end }}{{- /* End if $v.operatorGroup */}}
--- a/clustergroup/templates/core/operatorgroup.yaml
+++ b/clustergroup/templates/core/operatorgroup.yaml
@@ -21,15 +21,19 @@ kind: OperatorGroup
 metadata:
   name: {{ $k }}-operator-group
   namespace: {{ $k }}
+  {{- if (hasKey $v "targetNamespaces") }}
+    {{- if $v.targetNamespaces }}
 spec:
   targetNamespaces:
-    {{- if (hasKey $v "targetNamespaces") }}
       {{- range $v.targetNamespaces }}{{- /* We loop through the list of tergetnamespaces */}}
   - {{ . }}
       {{- end }}{{- /* End range targetNamespaces */}}
-    {{- else }}
+    {{- end }}{{- /* End if $v.targetNamespaces */}}
+  {{- else }}
+spec:
+  targetNamespaces:
   - {{ $k }}
-    {{- end }}{{- /* End of if operatorGroup */}}
+  {{- end }}{{- /* End of if (hasKey $v "targetNamespaces") */}}
   {{- end }}{{- /* range $k, $v := $ns */}}
   {{- end }}{{- /* End of if operatorGroup */}}
   {{- else if kindIs "string" $ns }}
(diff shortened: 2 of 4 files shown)


clustergroup/templates/plumbing/applications.yaml
```diff
@@ -78,7 +78,7 @@ spec:
             - name: global.namespace
               value: {{ $.Values.global.namespace }}
             - name: clusterGroup.name
-              value: {{ .Values.clusterGroup.name }}
+              value: {{ $.Values.clusterGroup.name }}
```
mbaldessari, line 81: Spurious change?

examples/values-example.yaml
```diff
@@ -90,6 +93,32 @@ clusterGroup:
         kind: ManagedClusterInfo
         jsonPointers:
         - /spec/loggingCA
+    appset:
```
mbaldessari, line 96: Spurious change?

tests/clustergroup-normal.expected.yaml
```diff
@@ -123,6 +123,32 @@ data:
           namespace: open-cluster-management
           path: common/acm
           project: datacenter
+        appset:
```
mbaldessari, line 126: Spurious change?

tests/clustergroup-normal.expected.yaml
```diff
@@ -689,75 +715,6 @@ status: {}
 # Source: clustergroup/templates/plumbing/applications.yaml
 apiVersion: argoproj.io/v1alpha1
 kind: Application
-metadata:
```
mbaldessari, line 692: Spurious change? I suppose this one got moved due to the above spurious changes?

clustergroup/templates/_helpers.tpl
```diff
@@ -202,13 +202,19 @@ kind: OperatorGroup
 metadata:
   name: {{ $k }}-operator-group
   namespace: {{ $k }}
+      {{- if (hasKey $v "targetNamespaces") }}
+      {{- if $v.targetNamespaces }}
```
strangiato, line 206: Isn't this functionally the same as `hasKey $v "targetNamespaces"`?

Do you need both?

clustergroup/templates/_helpers.tpl
```diff
@@ -202,13 +202,19 @@ kind: OperatorGroup
...
+      {{- if (len $v.targetNamespaces) }}
 spec:
   targetNamespaces:
-      {{- if (hasKey $v "targetNamespaces") }}
         {{- range $v.targetNamespaces }}{{- /* We loop through the list of tergetnamespaces */}}
   - {{ . }}
         {{- end }}{{- /* End range targetNamespaces */}}
```
mbaldessari, line 212: Can we fix the comments here? It's a bit confusing to understand which end refers to what

mbaldessari: Maybe we can also indent them to the left a bit (the if statements and the end statements) as it makes them a bit more readable

clustergroup/templates/core/operatorgroup.yaml
```diff
@@ -21,13 +21,17 @@ kind: OperatorGroup
...
   name: {{ $k }}-operator-group
   namespace: {{ $k }}
+    {{- if (hasKey $v "targetNamespaces") }}
+    {{- if $v.targetNamespaces }}
 spec:
   targetNamespaces:
-    {{- if (hasKey $v "targetNamespaces") }}
       {{- range $v.targetNamespaces }}{{- /* We loop through the list of tergetnamespaces */}}
   - {{ . }}
       {{- end }}{{- /* End range targetNamespaces */}}
```
mbaldessari, line 30: the end comment could use adjusting

clustergroup/templates/_helpers.tpl
claudiol: @strangiato The default behavior is that if you do not have `targetNamespaces` defined, but you have `operatorGroup: true`, then we will add the namespace to the OperatorGroup manifest like so:

```yaml
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  name: openshift-distributed-tracing-operator-group
  namespace: openshift-distributed-tracing
spec:
  targetNamespaces:
    - openshift-distributed-tracing
 ```
So in order for us to make it a global OperatorGroup you have to define both the `operatorGroup: true` and the `targetNamespaces: []` or `targetNamespaces: null`.  Hope that makes sense.

clustergroup/templates/plumbing/applications.yaml
claudiol: @mbaldessari In my tests this was giving me issues.  Perhaps we can talk about it this morning.

tests/clustergroup-normal.expected.yaml
claudiol: @mbaldessari Same here

clustergroup/templates/_helpers.tpl
claudiol: @mbaldessari I will clean this up.

clustergroup/templates/core/operatorgroup.yaml
claudiol: @mbaldessari Same cleanup here.

clustergroup/templates/_helpers.tpl
claudiol: Without the first if, that checks for null, the template explodes.
Error: error calling len: len of nil pointer

PR #541, opened 2024-06-10
Make resourcehealthchecks configurable

mbaldessari:
- **Make resourcehealthchecks configurable**
- **Update tests after upgrading resourceHealthChecks change**


beekhof:
@mbaldessari how does one drive this though?

PR #545, opened 2024-06-11
configure annotation based resource tracking

mbaldessari:


mbaldessari:
Moved to common from https://github.com/validatedpatterns/clustergroup-chart/pull/1

PR #560, opened 2024-07-18
Optionally allow for applications to be rolled out in an ordered fashion

beekhof:
Pure GitOps would have everything be infinitely declarative, however real life is messy and not all applications fit the declarative model.  Rather than push the retries higher and higher, allow knowledgable authors to optionally define an order that applications should be rolled out in.


mbaldessari:
I think this is a sensible change. I'd probably rename wave to syncwave or something like that to make it a bit more explicit. Also add it to values.json.schema and maybe in one example

beekhof:
Modified. @day0hero can you test that the comment is actually true?

soukron:
👀

PR #585, opened 2024-09-08
Use $group.name in clusterset metadata

mbaldessari:
Closes: validatedpatterns/regional-resiliency-pattern#5


mbaldessari:
@TomerFi Let me know if this makes sense or if you prefer to drop the change in the regional pattern. Thanks!

TomerFi:
@mbaldessari Yes - it makes perfect sense. Thank you!

mbaldessari:
Thanks!

PR #597, opened 2024-09-16
Factor out ansible into separate collection repo

mhjacks:


mhjacks:
WIP until all changes are made to the individual chart repos with ansible refs

mbaldessari:
We now released clustergroup 0.9.1 which points to the split-out imperative image, so I am merging this (will add some cautionary warnings here that you need the 0.9.* series)

PR #605, opened 2024-09-30
Fix scripts after common slim

darkdoc:
- **Update secret backend change script**
- **Update preview script**


mbaldessari:
This fixes at least some of the current brokenness. Merging this to `main` until we tackle the other corner cases

PR #608, opened 2024-10-07
Update Makefile to reflect a warning if no storageclass is present

### LINKED ISSUE #607: Do not enforce the existance of any storage class
UPI or bare metal clusters may not have any storage class.

As today, the validate-cluster target in the common/Makefile fails if there's no storage class.

Is there any reason to keep this requirement so strong in the Makefile?

soukron:
UPI or bare metal clusters may not have any storage class.

As today, the validate-cluster target in the common/Makefile fails if there's no storage class.

Is there any reason to keep this requirement so strong in the Makefile.

Fixes #607


mbaldessari:
The reason we had this is that people would deploy multicloud gitops on a cluster without default storage class and the vault would not come up, because they never specified a storage class for it.

I suppose downgrading it to a warning is a reasonable compromise.

Mergin this, thanks!

soukron:
Thanks! I wasn't entirely sure about making it a warning and/or adding the condition based on the vault installation.

PR #624, opened 2025-05-12
 Infer the git remote from the branch name

yuvalk:
this is better than assuming it's called `origin` because in many cases
 people want to use forks on the same local directory


mbaldessari (commented):
I like this. It is probably best to move the TARGET_REPO definition after TARGET_BRANCH and TARGET_ORIGIN. Makefile might be smart enough to resolve things properly, but it's best to be clearer.

PR #639, opened 2025-09-24
fix: Drop the quiet from the podman system connection list cmd

### LINKED ISSUE #637: Some podman versions don't have the -q switch when listing the connection list
```
podman system connection list -q 
```

Throws an error on the screen on most WSL installs. We need to catch that and either not display it or handle things properly

darkdoc:
The quiet option only available from podman 4.3. Instead just drop the
header line and count the connections manually. This way we would still
be able to see any other errors.

resolves #637


mbaldessari:
Thanks!

PR #654, opened 2026-03-27
Detect and run commands directly when already in a container

yukinchan:


mbaldessari:
I suspect you want to move the check after PATTERN_DISCONNECTED_HOME

yukinchan:
@mbaldessari good point.

